PR #71, opened 2022-10-07
Small fix to allow finch_exec to execute multiple top-level expressions

adadima:
finch_exec was failing for multiple lines of code because Meta.parse stops at new line and only considers one top level expression at a time, I just enclosed the input to Meta.parse in a `begin end` block.


willow-ahrens (approved):
Looks good! All tests pass, so I'll merge.

Issue #65, opened 2022-09-20
Code diffs fail because of Int64 vs Int32 problems


willow-ahrens:
I think the right approach here is to be specific about which int type we want to use for these tests, then enable the code diff tests again once that's fixed. Currently these tests need to be requested specifically. [see also](https://github.com/willow-ahrens/Finch.jl/blob/9dc3a188d8be8d666bfd292a1ddd34abf319c1d0/test/runtests.jl#L7)

willow-ahrens:
Fixed by #84

PR #84, opened 2022-12-08
Wma/reproducible

willow-ahrens:
A quick fix to make Finch reproducible, will be improved by #83


willow-ahrens:
Should also fix https://github.com/willow-ahrens/Finch.jl/issues/65

Issue #80, opened 2022-12-07
Delta Encoding Index Structures in Level Formats


willow-ahrens:

willow-ahrens:
closed by https://github.com/willow-ahrens/Finch.jl/pull/90

PR #90, opened 2023-01-02
Wma/deltas

willow-ahrens:
should improve test coverage


willow-ahrens:
closes https://github.com/willow-ahrens/Finch.jl/issues/80

Issue #94, opened 2023-01-11
Add implementations of == and isequal for Fibers.


willow-ahrens:
two arrays are `==` if their sizes are equal `size(A) == size(B)` and all of their elements are `==`. similar for `isequal`.

Let's make two implementations of these functions using Finch. We can use the implementation of [`Base.copyto!`](https://github.com/willow-ahrens/Finch.jl/blob/ddbdf3e6313bacb0119c937b1a8e780c1baa51a0/src/fibers_meta.jl#L33) as a template for how we can express the problem using Finch itself.

The overloads for `==` on `Fiber` types will look similar to those of [`StaticArrays` on `AbstractArray` types](https://github.com/JuliaArrays/StaticArrays.jl/blob/f7b8ed21c946fef2e9ecb00d7a27bccdd2623976/src/linalg.jl#L12).

willow-ahrens:
We'll also need tests that excercise all of the cases. For example the arrays could be equal, not equal, same size, different size, different array types (Array vs. Fiber)

willow-ahrens:
We might also want to test these on fibers with the same formats and different formats, just to be sure.

PR #98, opened 2023-01-23
Implement == on fibers

lylime:


src/fibers_meta.jl
```diff
@@ -30,6 +30,43 @@ function fiber(arr, default=zero(eltype(arr)))
     Base.copyto!(Fiber((DenseLevel^(ndims(arr)))(Element{default}())), arr)
 end
 
+@generated function Base.:(==)(A::Fiber, B::Fiber)
```
willow-ahrens, line 33: This looks great! Since the body of the generated function is the same for all three, perhaps we could add a helper function that all three overloads call? Something like `helper_equal(A, B)` that implements the generator. I should have done this for copyto! as well I think.

src/fibers_meta.jl
```diff
@@ -30,6 +30,43 @@ function fiber(arr, default=zero(eltype(arr)))
...
+    A = virtualize(:A, A, LowerJulia())
+    idxs = [Symbol(:i_, n) for n = getsites(A)]
+    idk = getsites(A)
+    return quote
+        check = Scalar(size(A) == size(B))
```
willow-ahrens, line 38: This is a small suggestion: The Julia compiler likely doesn't understand the contents of the `Scalar` type very well, so it may be better to implement this with the single line `size(A) == size(B) || return false` at the top, then the rest of the function below. If the sizes are known statically, Julia can then constant propagate and inline the function to false.

willow-ahrens (commented):
Great work! The test cases are really thorough and the implementation is just what we need! I requested a few very minor changes, and we can merge once those are fixed.

lylime:
Fixes #94

lylime:
Addressed above comments -- please let me know if you have any other feedback!

willow-ahrens:
Looks great! And there's a fix for copyto as well!

PR #103, opened 2023-02-05
Store tensors as program state, partially redesign Level Abstraction

willow-ahrens:


willow-ahrens:
TODOs for merging:

- [ ] clean up initializing vs. non-initializing constructors (add initialized Boolean or wrapper perhaps?)
- [ ] Split test reference into 32-bit and 64-bit
- [ ] test that we can print working reprs
- [ ] change constructor tests to print output and add macro fiber constructors
- [ ] change format tests to print output
- [ ] add window modifiers back

miguelraz:
So close... keep going! 💪🏾

willow-ahrens:
Thanks for the encouraging words! This PR was more complicated than I expected haha.

Issue #92, opened 2023-01-10
Row Major -> Column Major


willow-ahrens:
It's unfortunate, but Fibers would need to be column major to match other arrays in the language so that operations that require concordant iteration (like Broadcast) can be efficient by default.

willow-ahrens:
One potential problem resulting from this is that we can no longer read off the dimensions of a fiber from it's constructor:
`@fiber d(10, sl(20, e(0)))` used to be a 10x20 csr matrix, but would now be a 20x10 csc matrix (the order of modes is backwards). Column major was a mistake for Julia tbh.

willow-ahrens:
One way to fix the dimension problem is to add the sublevel as the first argument of level constructors. Then we have:
Fiber `d(sl(e(0.0), 20), 10)` which would be a 20 x 10 matrix. This would also make the constructors simpler.

willow-ahrens:
Column Major Pros:
- Pointwise operations are concordant by default
- `@fiber d(sl(e(0.0), 20), 10)` puts the format names first and dimensions second, analogous to `Array{T, N}(undef, 20, 10)` which puts initializer first and dimensions second
Column Major Cons:
- My research group will be confused because they are all row major
- The `vector of subarrays` analogy is less obvious. If `A[i, j, k]` is a column major vector of vectors, we have to write it `A[k][i, j]`, whereas if it was row major we write `A[i][j, k]`.

Of course we can always add permutation wrappers, etc to support one or the other but the question here is what is the default for the fiber type that we are introducing?

PR #107, opened 2023-02-11
Wma/col major

willow-ahrens:


willow-ahrens:
fixes https://github.com/willow-ahrens/Finch.jl/issues/92

Issue #115, opened 2023-02-20
SelectStyle requested within looplet nest and Out-of-order writes do not error when not supported


kylebd99:
I'm running into an error when trying to use a custom operator. Further, registering an unchanged algebra results in a different error, and registering an algebra which references the custom operator produces a different error. As a side note, before I updated from 0.2.0 to the current main branch, the second and third code samples produced the correct output.

For example, when I run this code:
```
function f(a, b, c)
    return a+b+c
end 
t = @fiber(sl(sl(e(0.0))))
A = dropdefaults(copyto!(@fiber(sl(sl(e(0.0)))), SparseMatrixCSC([0 0 0 0; -1 -1 -1 -1; -2 -2 -2 -2; -3 -3 -3 -3])))
@finch @loop i j t[i, j] = f(A[i,j], A[i,j], A[i,j])
```
I get the following error:
```
RewriteTools.RuleRewriteError(call(~f, ~a...) => if isliteral(f) && (all(isliteral, a) && length(a) >= 1)
        #= /home/kylebd99/.julia/packages/Finch/De0AF/src/annihilate.jl:155 =#
        literal((getvalue(f))(getvalue.(a)...))
    end, Finch.IndexNotation.IndexNode(Finch.IndexNotation.call, nothing, nothing, Finch.IndexNotation.IndexNode[Finch.IndexNotation.IndexNode(Finch.IndexNotation.literal, f, nothing, Finch.IndexNotation.IndexNode[]), Finch.IndexNotation.IndexNode(Finch.IndexNotation.literal, 0.0, nothing, Finch.IndexNotation.IndexNode[]), Finch.IndexNotation.IndexNode(Finch.IndexNotation.literal, 0.0, nothing, Finch.IndexNotation.IndexNode[]), Finch.IndexNotation.IndexNode(Finch.IndexNotation.literal, 0.0, nothing, Finch.IndexNotation.IndexNode[])]))
```
However, if I run the following code:
```
function f(a::Float64, b::Float64, c::Float64)
    return a+b+c
end 
struct MyAlgebra <: Finch.AbstractAlgebra end
Finch.register(MyAlgebra)
t = @fiber(sl(sl(e(0.0))))
A = dropdefaults(copyto!(@fiber(sl(sl(e(0.0)))), SparseMatrixCSC([0 0 0 0; -1 -1 -1 -1; -2 -2 -2 -2; -3 -3 -3 -3])))
[...]

PR #116, opened 2023-02-21
Wma/check inorder

willow-ahrens:
Check that fibers are in order when accessing


willow-ahrens:
addresses https://github.com/willow-ahrens/Finch.jl/issues/115

Issue #120, opened 2023-03-03
Preliminary binary file format output


willow-ahrens:
Write binary file format output in a simple format with HDF5, then perhaps attempt to follow the binsparse-spec.

http://api.csswg.org/bikeshed/?url=https://raw.githubusercontent.com/GraphBLAS/binsparse-specification/7e62e7d77df5c49777d952784dad389a25e92347/spec/latest/index.bs#references

willow-ahrens:
Consider also https://github.com/GraphBLAS/binsparse-specification/issues/2 and https://juliaio.github.io/HDF5.jl/stable/

willow-ahrens:
This got merged, yay! Let's file new issues to keep the implementation up to date.

PR #126, opened 2023-03-03
add hdf5 io

willow-ahrens:
This PR contains the Binsparse binary format (using hdf5), as well as a finch-specific binary format. Currently, the fiber is written in column-major order and is 1-indexed.

TODO:

- [x] Add MatrixMarket, TensorMarket, TNS.
- [x] Add a section to the docs for different file formats


willow-ahrens:
addresses https://github.com/willow-ahrens/Finch.jl/issues/120

PR #128, opened 2023-03-05
Bellmanford

Cece2520:
Pull request to review current progress.

Moved graph functions to Apps folder and added testing for those functions.


willow-ahrens (changes requested):
This looks great, and I'd like to get it merged! Before we do, could you include the apps.jl file into the benchmarks.jl file, and call the apps functions directly from the benchmarks to reduce code duplication?

Issue #99, opened 2023-01-23
0 Based Internal Vectors for Fibers


Wimmerer:
0-based languages will have sparse arrays represented internally by 0 based vectors. While Fibers should always use `getindex` starting with 1, support for internal 0 based representation is key for workloads that are, for instance, solver-centric.

willow-ahrens:
I agree we should implement this, there's a question of how much parameterization we should expose.

Most of the levels are parameterized with a `Tp` argument, specifying the integer type of positions in the fiber. However, the user probably doesn't want to change this property on a per-level basis (it would cause type instability). We can default Tp to be the same position type as that used by the sublevel (it currently just defaults to `Int`, so you have to set it for every level if you want to deviate).

The next challenge is of course 0 or 1 indexing. We can change the starting position to 0 with the same mechanism as above, no problem. However, for indices it's a different story. People might want to index their tensor with BigInt starting at -7, a different problem than the kind of "starting index isn't 1" that we talk about here. When we store to a binary format or reinterpret for another language, we do need to shift the stored values of the indices. Perhaps the answer to this problem is more granular, and should involve interpreting 0-indexed arrays as 0-indexed _in julia_, then shifting them by one _in julia_ as well. I'm open to other suggestions.

willow-ahrens:
so the operation to read a 0-indexed vector from memory would look like
```
A_raw = fbrread("A.fbr")
A_raw::Fiber{SparseListLevel{Int, 0, ElementLevel{0.0, Float64, 0}}}
A = Fiber!(SparseListLevel{Int, 1, ElementLevel{0.0, Float64, 1}}}
@finch @loop j i A[i] = A_raw[i-1]
```

willow-ahrens:
Alternately, we can say that fibers always start at `one(Ti)` or `zero(Ti)`, and this is determined by whether the positions start at one or zero as well.

Wimmerer:
I currently have a `Bi` parameter which isn't really constrained but is assumed to be 1 or 0. So that seems to match up well?

willow-ahrens:
Two things need to work:1. Fibers indexed from 1, 0, or -72. Fibers indexed from 1 that subtract one internally (storage looks 0 indexed, semantics look 1 indexed)I am trying to figure out if I can do both at once or if these need to be separate concepts.On Mar 9, 2023, at 12:01 PM, Will Kimmerer ***@***.***> wrote:﻿
[...]

PR #139, opened 2023-03-10
Wma/cindex

willow-ahrens:
Add zero-based indexing for Fibers


willow-ahrens:
fix https://github.com/willow-ahrens/Finch.jl/issues/99

PR #176, opened 2023-04-26
Wma/precompile and docs

willow-ahrens:
This PR is to finalize the compilation story with Finch. Writing compilers in Julia gets tricky when we want the user to interact with the compiler.

Julia uses a "world age" to describe the set of defined functions at a point in time. Generated functions run in the same world age in which they were defined, so they can't call functions defined after the generated function. This means that if Finch uses generated functions, users can't define their own functions without redefining the generated function.

This used to be automated by the `Finch.register` function that just evals the generator once again. However, this is not composable when multiple packages use Finch, so we began associating different worlds with different algebras. The idea was that different users could use different algebras for different sets of functions. Unfortunately, since Finch gets called from methods like Broadcast and stuff, it quickly became too complicated to automatically figure out which algebra was appropriate top evaluate the types in a broadcast.

~~The new approach is to use RuntimeGeneratedFunctions, which free up the world age. This does mean that every dynamic finch call will need a dictionary lookup and a dynamic function invocation~~ < Turns out this approach had unacceptable overhead. Instead we figured out a way to call the generator in the latest world age, and left behind a trail of breadcrumbs in case anyone wanted to invalidate finch generated functions manually (with `refresh()`).

If we're careful, users probably don't need to ever call `refresh()` if everyone designs things so that they define all relevant properties before calling each kernel.

It's a compromise, but I think it delivers as much as possible on the idea of "Jitted tensor compiler in Julia".


willow-ahrens:
Ping https://discourse.julialang.org/t/constant-propagation-vs-generated-functions/83788 when this is finished to relay my experience.

willow-ahrens:
TODO on this before merging:

- [ ] Change other instances of generated functions that call finch functions to also be world-age safe
- [ ] Add docs and tests about precompiling Finch functions
- [ ] add some docs about typical finch usage.

willow-ahrens:
Also consider a ping to https://github.com/NHDaly/StagedFunctions.jl/blob/master/src/StagedFunctions.jl#L116, https://github.com/FluxML/Zygote.jl/issues/22, and RuntimeGeneratedFunctions?

PR #197, opened 2023-05-16
Fix bug with is_laminable_updater in Sparse Triangle Level

radha-patel:


willow-ahrens (approved):
lgtm! added the reproducer you mentioned

PR #204, opened 2023-05-19
Partial lowering and introspection interface

willow-ahrens:
@wraith1995 
The relevant invocation was
```
using Finch
y = @fiber d(e(0.0))
A = @fiber d(sl(e(0.0)))
x = @fiber sl(e(0.0))
for i = 1:10
    Finch.execute_code(:ex, typeof(Finch.@finch_program_instance begin
        @loop j i y[i] += A[i, j] * x[j]
    end), Finch.DefaultAlgebra(), Finch.WillResumeContext(Finch.LowerJulia(), i)) |> Finch.striplines |> Finch.unblock |> display
    println("\n\n\n")
end

### CODE REVIEW

**Comment by willow-ahrens** [comment]
an idea: you could totally do a [Cthulhu](https://github.com/JuliaDebug/Cthulhu.jl)-style interface to allow you to select which resumable you want to lower.

**Comment by willow-ahrens** [comment]
oh and wouldn't it be cool to also be able to schedule subprograms through the interface?

**Comment by wraith1995** [comment]
Right now, I think the key things are:

1. Select by pass
2. Select by location
3. Select by custom filter function

A nicer interactive interface would be great. 

I think that scheduling subprograms through something like this is pretty important - I think we need a nice way to see what the commands are actually doing because otherwise it is rather painful to think about. Cthulhu might be a good metaphor here because IIRC they have some notion of going back in time or something?

**Comment by willow-ahrens** [comment]
Heads up: I'd like to merge a few interface changes to the `execute_code` function today (I want it to use kwargs).

File: src/resumables.jl [hunk 1]
```diff
@@ -0,0 +1,319 @@
...
+
+
+function (ctx::DebugContext)(node, style)
+    println("On:", ctx.control)
+    display(node)
+    println("style:", style)
+    if node isa Resumable
+        if should_resume(ctx.control, ctx, node.root, style, node.meta)
+            println("resuming...")
+            control = evolve_control(ctx.control, ctx.ctx, node.root, node.style)
```
willow-ahrens, line 151: why is there ctx.ctx here?

src/resumables.jl
```diff
@@ -0,0 +1,319 @@
...
+function (ctx::DebugContext)(code:: Expr)
+    if iscompiled(code)
+        return code
+    end
+    println("Through...")
+    Prewalk(node -> 
+    if node isa Resumable 
+        println("trying...")
+        display(node)
+        ret = (ctx::DebugContext)(node, node.style)
```
willow-ahrens, line 185:
```
DebugContext(control, node.ctx)(node.body, node.style)
```

src/resumables.jl
```diff
@@ -0,0 +1,319 @@
...
+        nxt = lower(node, DebugContext(ctx.ctx, control), style)
+        println("Finished going....")
+        display(node)
+        println("to....")
+        display(unblock(striplines(nxt)))
+        return nxt
+    end
+end
+
+function (ctx::DebugContext)(code:: Expr)
```
willow-ahrens, line 176: function resume_lowering

willow-ahrens:
I think we can make this work, actually. The only statements that modify the context with a strong sequential dependency are the declare/freeze/thaw/forget statements. If we enforce that sequences cannot be nested, then we can handle e.g. `sequence(a, b, declare(tensor_A), c, d)` as `sequence(a, sequence(b, sequence(declare(tensor_A), sequence(c, d)))))`. If we process `sequence(declare(tensor_A), body)` as an atomic lowering unit (`tensor_A` would only be declared in the context used to lower `body`), then there would be no more stateful context dependencies. 

All of this breaks if we want to lower tensors with static information. We don't currently support this, but imagine asking the compiler to unroll a loop and perform some of the kernel at compile time.

willow-ahrens:
Another point here is the modifications to preamble and epilogue. These are stateful because they have side effects (modifying preamble/epilogue will sometimes update a different context than the one we are currently lowering). If we lower a context and halt execution of the body, then emit it's preamble or epilogue, then lower the body, we might be in trouble because updates to the preamble or epilogue produced during lowering of the body might be missed. One approach to this could be to handle preambles and epilogues as a post-processing pass. Due to how often we can (and should) call contain, another approach might be to handle preamble and epilogue as greedily as possible, avoiding the problem altogether.

willow-ahrens:
Apart from the above two concerns (which actually feel very solveable), I think that we can make this a functional compiler and also make this debugger function the way you want it to (i.e. choosing which phase of a pipeline to lower, etc.)

willow-ahrens:
Yeah, I think that's what I meant. Roughly, we could emit code that looks like:
```
contain(for i = 1:10
contain(A[i] += thunk(B[i] * C[j], epilogue = :(j += 1))
end)
```

willow-ahrens:
And we could lower it later. But it would sorta defeat the purpose of your debugger because it's so unreadable. I think the actual answer might be that all nontrivial uses of preamble and epilogue can be handled in a single call to the context.

wraith1995 (approved):

Issue #181, opened 2023-05-02
Generalize integration


willow-ahrens:
Integration rules are currently special cased on the loop body ast. For context, strength reduction rules are rules of the form `for i = 1:n A[] += 1 end => A[] += n`. These rules are special cased because the rule needs to match the loop index and also needs to match the increment expression. There's also some tricky pieces since we can't always remove the loop entirely if the loop contains other expressions that cannot be strength reduced. My suggested fix is to match once on the loop expression `for ~i = ~ext; ~body end`, then inside of the rule perform a postwalk over body, replacing `A[j...] += x` with `if diag[j, getstop(ext)] A[j...] += x * measure(ext) end` where `j` and `x` are orthogonal to `i`. This ensures that the update only happens once, and allows for cases where `x` depends on a variable computed inside an inner loop. The generated code will be better if you can use a spike instead of a diagonal mask. Also, if we want to support this in it's full generality, we need to generalize our `orthogonality` calculation to propagate through constants.

willow-ahrens:
Closed in #250. Thx Jaeyeon!

PR #217, opened 2023-05-30
add continuous level

nullplay:
We want to add interval list as a new level.
This will be the first step before we move on dealing with closeness/openness of intervals.

## Todo list
- [x] Supporting Sparse Output (get_updater)
- [x] Adding Test Cases for continuouslevel.
- [x] Generalized Strength Reduction Rule (#181) on nested for loops.


nullplay:
While I was adding test cases in test_conversion.jl, I met some errors I cannot understand.

The error message when running `test Finch` on Pkg.
```
[ Info: Testing Fiber Conversions
convert 0 @fiber(cont(e(false))) 1: Error During Test at /Users/jaeyeonwon/Finch.jl/test/test_conversions.jl:44
  Got exception outside of a @test
  BoundsError: attempt to access 0-element Vector{Int64} at index [0]
  Stacktrace:
    [1] getindex
      @ ./essentials.jl:13 [inlined]
    [2] macro expansion
      @ ~/Finch.jl/src/lower.jl:213 [inlined]
    [3] macro expansion
      @ ~/Finch.jl/src/execute.jl:6 [inlined]
    [4] macro expansion
      @ ~/Finch.jl/src/util.jl:15 [inlined]
    [5] execute(ex::Finch.FinchNotation.LoopInstance{Finch.FinchNotation.IndexInstance{:i}, Finch.FinchNotation.IndexInstance{:(:)}, Finch.FinchNotation.AssignInstance{Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:check, Scalar{true, Bool}}, Finch.FinchNotation.UpdaterInstance{Finch.FinchNotation.CreateInstance}, Tuple{}}, Finch.FinchNotation.VariableInstance{:&, Finch.FinchNotation.LiteralInstance{&}}, Finch.FinchNotation.CallInstance{Finch.FinchNotation.VariableInstance{:(==), Finch.FinchNotation.LiteralInstance{==}}, Tuple{Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:tmp, Fiber{ContinuousLevel{Int64, Int64, ElementLevel{false, Bool}}}}, Finch.FinchNotation.ReaderInstance, Tuple{Finch.FinchNotation.IndexInstance{:i}}}, Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:ref, Fiber{SparseListLevel{Int64, Int64, ElementLevel{false, Bool}}}}, Finch.FinchNotation.ReaderInstance, Tuple{Finch.FinchNotation.IndexInstance{:i}}}}}}}, a::Finch.DefaultAlgebra)
      @ Finch ~/Finch.jl/src/util.jl:15
    [6] execute(ex::Finch.FinchNotation.LoopInstance{Finch.FinchNotation.IndexInstance{:i}, Finch.FinchNotation.IndexInstance{:(:)}, Finch.FinchNotation.AssignInstance{Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:check, Scalar{true, Bool}}, Finch.FinchNotation.UpdaterInstance{Finch.FinchNotation.CreateInstance}, Tuple{}}, Finch.FinchNotation.VariableInstance{:&, Finch.FinchNotation.LiteralInstance{&}}, Finch.FinchNotation.CallInstance{Finch.FinchNotation.VariableInstance{:(==), Finch.FinchNotation.LiteralInstance{==}}, Tuple{Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:tmp, Fiber{ContinuousLevel{Int64, Int64, ElementLevel{false, Bool}}}}, Finch.FinchNotation.ReaderInstance, Tuple{Finch.FinchNotation.IndexInstance{:i}}}, Finch.FinchNotation.AccessInstance{Finch.FinchNotation.VariableInstance{:ref, Fiber{SparseListLevel{Int64, Int64, ElementLevel{false, Bool}}}}, Finch.FinchNotation.ReaderInstance, Tuple{Finch.FinchNotation.IndexInstance{:i}}}}}}})
      @ Finch ~/Finch.jl/src/execute.jl:1
    [7] macro expansion
      @ ~/Finch.jl/src/execute.jl:97 [inlined]
    [8] macro expansion
      @ ~/Finch.jl/test/test_conversions.jl:47 [inlined]
    [9] macro expansion
      @ ~/.julia/juliaup/julia-1.9.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
   [10] macro expansion
      @ ~/Finch.jl/test/test_conversions.jl:45 [inlined]
   [11] macro expansion
      @ ~/.julia/juliaup/julia-1.9.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
   [12] top-level scope
      @ ~/Finch.jl/test/test_conversions.jl:2
   [13] include(fname::String)
      @ Base.MainInclude ./client.jl:478
   [14] macro expansion
      @ ~/Finch.jl/test/runtests.jl:85 [inlined]
   [15] macro expansion
      @ ~/.julia/juliaup/julia-1.9.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
   [16] top-level scope
      @ ~/Finch.jl/test/runtests.jl:81
   [17] include(fname::String)
      @ Base.MainInclude ./client.jl:478
   [18] top-level scope
      @ none:6
   [19] eval
      @ ./boot.jl:370 [inlined]
   [20] exec_options(opts::Base.JLOptions)
      @ Base ./client.jl:280
   [21] _start()
      @ Base ./client.jl:522
```

Region of interest in `test_conversion.jl`:
```
           for base in [
               Element{false},
           ]
               if true
                   for inner in [
                       () -> Continuous(base()),
                   ]
                       for arr in [
                           fill(false, 5),
                          ]
                           ref = @fiber sl(e(false))
                           res = @fiber sl(e(false))
                           ref = dropdefaults!(ref, arr)
                           tmp = Fiber!(inner())
                           begin
                               @finch (tmp .= 0; @loop i tmp[i] = ref[i])
                               check = Scalar(true)
                               @finch @loop i check[] &= tmp[i] == ref[i]
                               println(check[])
                           end
                       end
                   end
               end
           end
```

#### Problem : While there is an error when running test using Pkg, running above test code on REPL works well without an error.

willow-ahrens:
This is likely caused by out of bounds array access. Julia only sometimes checks bounds, because it's expensive to check every access (in C, this would be segfault). You can run `julia --check-bounds=yes` if you want Julia to error on every out of bounds access.

willow-ahrens:
As a followup, try to print `@finch_code` for the first one that errors.

nullplay:
I was planning to merge this interval list level to main. Do you want me to maintain this level in separate branch?

nullplay:
@willow-ahrens  Can you review this level format?

Basically, it is an interval list level with read/update support along with some test cases.

I solved a strength reduction problem #181 (which corresponds to integral reduction on piece-wise constant function) on  nested loop by adding rewriting in default style instead of dealing in symbolic simplifier. Check [src/lower.jl](https://github.com/willow-ahrens/Finch.jl/blob/cc0561454ad39dd620625593c997d3a525fc05db/src/lower.jl#L254)

src/tensors/levels/continuouslevels.jl
```diff
@@ -0,0 +1,314 @@
...
+    Tp = lvl.Tp
+    Ti = lvl.Ti
+    qos = ctx.freshen(tag, :_qos)
+    qos_fill = lvl.qos_fill
+    qos_stop = lvl.qos_stop
+    dirty = ctx.freshen(tag, :dirty)
+    
+    Furlable(
+        tight = lvl, # Jaeyeon : Something related to error handling in unfurl.jl
+        size = virtual_level_size(lvl, ctx), # Jaeyeon : list of all extents of subfiber's shape?
```
willow-ahrens, line 279: This is right. Also, I'll be removing the size field on furlable in subsequent versions of the code.

src/lower.jl
```diff
@@ -250,6 +250,18 @@ function lower(root::FinchNode, ctx::AbstractCompiler, ::DefaultStyle)
     elseif root.kind === loop
         @assert root.idx.kind === index
         idx_sym = ctx.freshen(root.idx.name)
+        
+        # Generalized Strength Reduction
```
willow-ahrens, line 254: This is not strictly correct, it's too eager. There might be more than one access, just because one of them can be reduced does not mean that all of them can.

willow-ahrens: e.g.
```
for i = _
    A[] += B[i]
    C[] += 1
end
```

willow-ahrens: My approach here would be to instead guard the assignment with a `diagmask`.

```
for i = _
    A[] += B[i]
    if diagmask[i, getstop(root.ext)]
        C[] += measure(root.ext)
    end
end
```

willow-ahrens: Then you could also add a few rules about collapsing identical conditions, and perhaps we could make a mask called `once[i]` which is just a `spike(false, true)`

willow-ahrens: Could we perhaps separate issue #181 from this PR?

willow-ahrens (changes requested):
This looks great! Can we rename `Continuous` to `SparseRLE` or similar? I think this level is a separate concept from a continuous index. If you rename make a separate PR for the #181 changes, I think we can merge this one.

src/lower.jl
nullplay: Yea I agree that it would be better handled in separate PR. I'll revert lower.jl.

nullplay:
Continuous -> SparseRLE

willow-ahrens:
Is the following behavior intended?

```
julia> x = @fiber(srl(e(0.0)), [0, 0, 1, 1, 1, 0, 0])
SparseRLE (0.0) [1:7]
├─[3:3]: 1.0
├─[4:4]: 1.0
├─[5:5]: 1.0
```

willow-ahrens (approved):

willow-ahrens (commented):
Should the SparseRLE merge adjacent runs by default?

nullplay:
> Is the following behavior intended?
> 
> ```
> julia> x = @fiber(srl(e(0.0)), [0, 0, 1, 1, 1, 0, 0])
> SparseRLE (0.0) [1:7]
> ├─[3:3]: 1.0
> ├─[4:4]: 1.0
> ├─[5:5]: 1.0
> ```

This was intended after some consideration. Because I want to view integer tensors as a pinpoint coordinate ([a,a]) in our continuous indices, an automatic merging (A[3:5] = 1) would semantically result in different tensor in continuous domain.

nullplay:
I think you've raised a very interesting question: How should we interpret the expression `A[2:4]=1` in the context of a continuous tensor field? We have previously discussed two options for interpreting existing integer tensors in a field: (1) a pinpoint interval (`[n,n]`) and (2) the interval with length 1 (`[n, n+1)`). So far, I believe that interpreting the existing tensor as a pinpoint field is the correct approach.

However, I have a concern regarding the syntax `A[2:4]=1`. It seems like this syntax would insert a single piece-wise constant value of 1 into the interval [2,4) in the field A. However, from a semantic standpoint, is `A[2:4]=1` actually inserting three piece-wise constants into three discontinuous singleton intervals (`[2,2]`, `[3,3]`, `[4,4]`) in the integer world? If that is the case, should we consider introducing a different syntax to represent the insertion of a single(`[2,4)`) piece-wise constant value?

willow-ahrens:
I think it's okay to have discrete runs: A[2:4] = 7 means that all of the _integers_ between 2 and 4 are 7, and B[2.0:4.0] = 7 means that all the _reals_ between 2.0 and 4.0 are 7. `C[i] += A[i] * B[i]` is undefined because we never specified the value of `A` at e.g. 2.5. Thus, we should say either `C[i] += coalesce(A[i], missing) * B[i]` or `C[i] += A[floor(i)] * B[i]` to request the appropriate behavior.

willow-ahrens:
Like, I think that a tensor with integer index types should be understood as defined only on integers, same for floats/reals

PR #219, opened 2023-06-02
Add a type to represent open/closed interval endpoints at runtime

willow-ahrens:
Here's an initial implementation of the `Limit` type that represents open/closed intervals at runtime. Note that when we are dealing with symbolic open/closed representations, we can use rewrite rules in [analyze_bounds.jl](https://github.com/willow-ahrens/Finch.jl/blob/main/src/symbolic/analyze_bounds.jl) that operate on the `limit` function, e.g. 

`call(limit, ~x, Inf) + call(limit, ~y, Inf) => call(limit, call(+, x, y), Inf)`

and the like


willow-ahrens:
@nullplay

willow-ahrens:
I used a float value for the sign because it encoded the semantics we needed surrounding cancellation (i.e. 1+eps + 1-eps = 2 +- eps). We really shouldn't ever need to add right open + left open, but this type allows us to detect when we do.

willow-ahrens:
Summarizing discussion today:
1. It's really nice for (1-eps) + eps = 1 (when we need to figure out what the number after e.g. prev_i is)
2. It's also the case that symbolically, we need 1 + eps + eps = 1 + eps (e.g. 2-eps + 4-eps = 6-eps)
3. Also, we need 2 * (1+eps) = 2+eps for scaled bounds
4. This means that a < b  is not equivalent to b == max(a + eps, b) (consider 1+eps < 1+eps)
5. However, somehow, a < b  is equivalent to b - a == max(0+eps, b - a).
6. I think that we need to figure out rules for max of limits anyway.

willow-ahrens:
Several points were raised in our discussion today, but I think that we have identified two conflicting uses of the `+` function for the limit type. 

1. We want to say `i = i_prev + eps` to get the matching bracket (closed or open) for the next interval. This would happen in e.g. https://github.com/willow-ahrens/Finch.jl/blob/c3bb2fbd4c6e09544dd38fcb5ea5d1b319cdb094/src/looplets/phases.jl#L80 This requires that `(x - eps) + eps = x`
2. We want `x::Limit + y::Limit` to add `x.val` and `y.val`, but propagate whether this limit is a left, right, or exact value. This requires that two right limits sum to another right limit `x + eps + eps = x + eps`.

The combination of these two behaviors mean that + is no longer associative. Consider the sum `eps + eps + -eps + -eps`. This could be any of:
`(eps + (eps + (-eps + -eps))) = eps`
`(eps + (eps + -eps)) + -eps = 0`
`((eps + eps) + -eps) + -eps = -eps`
In the end, we could only say that + is only associative if no `eps` cancellation occurs. The current implementation is always associative because `NaN` is produced in the event of cancellation.

We can guarantee that limits are cancellation-free during rewriting because `measure = getstop-getstart` will never subtract e.g. a right bracket from a right bracket. 

Our discussion also identified that supporting `a < b ` as `b == max(a + eps, b)` would be a problem under behavior 2. However, I think we don't need to support `<` because all calls to `query` compare measures with literal 0 which could be replaced with 0+eps as needed.

willow-ahrens:
@nullplay I'm nervous about modifying the index type for continuous/discrete, since indices are treated as symbols (it may be burdensome for an index to take arguments). Could you modify the `Extent` type for this purpose instead?

willow-ahrens:
i.e. could we just use `Ti` to indicate continuous/discrete? an example would be here: https://github.com/willow-ahrens/Finch.jl/blob/c3bb2fbd4c6e09544dd38fcb5ea5d1b319cdb094/src/tensors/levels/repeatrlelevels.jl#L26

willow-ahrens:
then, we could define

```
is_discrete(::Type{<:Integer}) = true
is_discrete(::Type{Float64}) = false
is_discrete(::Type{<:Fixpoint{<:Integer}}) = false
```
or something like that?

willow-ahrens:
Then, when you declare the size of a tensor, you could return a `ContinuousExtent` depending on whether `is_discrete(Ti)`. https://github.com/willow-ahrens/Finch.jl/blob/c3bb2fbd4c6e09544dd38fcb5ea5d1b319cdb094/src/tensors/levels/denselevels.jl#L120

willow-ahrens:
Then, you can define promotion rules for `ContinuousExtent`, e.g. https://github.com/willow-ahrens/Finch.jl/blob/c3bb2fbd4c6e09544dd38fcb5ea5d1b319cdb094/src/dimensionalize.jl#L231, e.g.

```
combinedim(ctx, a::Extent, b::ContinuousExtent) =
    ConinuousExtent(
        start = checklim(ctx, a.start, b.start),
        stop = checklim(ctx, a.stop, b.stop)
    )
```

willow-ahrens:
Then, right after this line (https://github.com/willow-ahrens/Finch.jl/blob/07ebad5b3902925df5d19d4376f0bd32eb9d5202/src/looplets/lookups.jl#L23C26-L23C26), I would assert that the loop dimension is discrete:

willow-ahrens:
My proposed changes would allow you to do continuous ranges without needing to modify Finch AST. Is there a reason we would want to modify Finch AST?

nullplay:
Before this branch becomes a huge change, I want to merge this into the main branch.

## Summary of Changes in this Pull Request:

1. Limit type (Epsilon) to represent open/closed endpoints of the interval.
2. ContinuousExtent and accompanying looplet lowering.
   - Currently, only SparseListLevel and SparseRLELevel support continuous iteration.
3. Fix symbolic interval analysis.

## Next Planned Steps:

1. Adding two types of reduction (counting and Lebesgue measure with `d(i)`).
2. Adding test cases.
3. Adding a new looplet; regular stepper.
4. Adding a new looplet; iteration over an interval list that has multiple overlapping intervals.

src/symbolic/analyze_bounds.jl
```diff
@@ -68,15 +68,15 @@ function get_bounds_rules(alg, shash)
         end),
 
         (@rule call(max, ~a1..., call(min, ~a2...), ~a3..., call(min, ~a4...), ~a5...) => if !(isdisjoint(a2, a4))
-            call(max, a1..., call(min, union(a2, a4)..., call(max, call(min, setdiff(a2, a4)...), call(min, setdiff(a4, a2)...)), a3..., a5...))
+            call(max, a1..., call(min, intersect(a2, a4)..., call(max, call(min, setdiff(a2, a4)...), call(min, setdiff(a4, a2)...)), a3..., a5...))
```
nullplay, line 71: Upon discovering potential inaccuracies in certain cases, I have switched the rule from `union` to `intersect`.

src/looplets/lookups.jl
```diff
@@ -37,7 +37,9 @@ function lower(root::FinchNode, ctx::AbstractCompiler,  ::LookupStyle)
             open_scope(body_3, ctx_2)
         end
         @assert isvirtual(root.ext)
-        if query(call(==, measure(root.ext.val), 1), ctx)
+
+        target = is_continuous_extent(root.ext) ? 0 : 1
+        if query(call(==, measure(root.ext.val), target), ctx)
```
nullplay, line 42: `measure(ext)` is defined differently depending on the type of `ext`. 

```
measure(ext::Extent) = call(+, call(-, ext.stop, ext.start), 1)
measure(ext::ContinuousExtent) = call(-, ext.stop, ext.start) 
```

nullplay (approved):

src/looplets/lookups.jl
willow-ahrens: does this make sense? I sortof think it does, but it's different than what we thought.

src/symbolic/analyze_bounds.jl
willow-ahrens: Okay, cool!

src/symbolic/analyze_bounds.jl
```diff
@@ -68,15 +68,15 @@ function get_bounds_rules(alg, shash)
...
         end),
 
         (@rule call(min, ~a1..., call(max, ~a2...), ~a3..., call(max, ~a4...), ~a5...) => if !(isdisjoint(a2, a4))
-            call(min, a1..., call(max, union(a2, a4)..., call(min, call(max, setdiff(a2, a4)...), call(max, setdiff(a4, a2)...)), a3..., a5...))
+            call(min, a1..., call(max, intersect(a2, a4)..., call(min, call(max, setdiff(a2, a4)...), call(max, setdiff(a4, a2)...)), a3..., a5...))
```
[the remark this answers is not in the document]
willow-ahrens: nice catch!

src/symbolic/analyze_bounds.jl
```diff
@@ -68,15 +68,15 @@ function get_bounds_rules(alg, shash)
...
         end),
 
-        (@rule call(min, ~a1..., call(max), ~a2...) => call(min, a1..., a2...)),
-        (@rule call(max, ~a1..., call(min), ~a2...) => call(min, a1..., a2...)),
+        (@rule call(min, ~a1..., call(max), ~a2...) => begin call(min, a1..., a2...) end),
```
willow-ahrens, line 78: why do we need a block here?

src/symbolic/analyze_bounds.jl
```diff
@@ -102,6 +102,15 @@ function get_bounds_rules(alg, shash)
 
         (@rule call(~f::isinvolution(alg), call(~f, ~a)) => a),
 
+        # Clamping rules
```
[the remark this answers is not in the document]
willow-ahrens: This rule is expensive but I don't know of a better option right now.

src/tensors/levels/sparserlelevels.jl
```diff
@@ -233,12 +237,12 @@ function instantiate_reader(fbr::VirtualSubFiber{VirtualSparseRLELevel}, ctx, ::
                                 body = (ctx, ext) -> Thunk( 
                                      body = (ctx) -> Pipeline([
                                         Phase(
-                                            stop = (ctx, ext) -> call(-, value(my_i_start), 1),
+                                            stop = (ctx, ext) -> call(-, value(my_i_start), is_continuous_extent(ext) ? Eps : Ti(1)),
```
willow-ahrens, line 240: Can you make this a function like `get_after(ext)` which specializes on `ContinuousExtent`? The boolean looks not very extensible.

willow-ahrens: or perhaps `get_eps(ext::Extent) = literal(1)` and `get_eps(ext::ContinuousExtent) = Eps`?

src/transforms/dimensionalize.jl
```diff
@@ -130,55 +130,63 @@ combinedim(ctx, a::NoDimension, b) = b
     stop
 end
 
+@kwdef struct ContinuousExtent
+    start
+    stop
+end
+
+is_continuous_extent(x) = false # generic
```
willow-ahrens, line 138: I feel like any place where we use this should probably be converted to a named function that specializes on the type of the extent

willow-ahrens: e.g. `measure` or `get_eps`

src/transforms/dimensionalize.jl
```diff
@@ -284,9 +305,16 @@ function combinedim(ctx, a::Widen{<:Extent}, b::Widen{<:Extent})
         stop = @f(max($(getstop(a)), $(getstop(b))))
     ))
 end
+function combinedim(ctx, a::Widen{<:ContinuousExtent}, b::Widen{<:ContinuousExtent})
+    Widen(ContinuousExtent(
+        start = @f(min($(getstart(a)), $(getstart(b)))),
+        stop = @f(max($(getstop(a)), $(getstop(b))))
+    ))
+end
+combinedim(ctx, a::Widen{<:Extent}, b::Widen{<:ContinuousExtent}) = throw(ArgumentError("Extent and ContinuousExtent cannot interact"))
```
willow-ahrens, line 314: perhaps you could add `... yet.`

test/reference64/typical_merge_twofinger.txt
```diff
@@ -41,48 +41,46 @@ quote
...
-                        z_lvl_qos_stop = max(z_lvl_qos_stop << 1, 1)
-                        Finch.resize_if_smaller!(z_lvl.idx, z_lvl_qos_stop)
-                        Finch.resize_if_smaller!(z_lvl_2.val, z_lvl_qos_stop)
-                        Finch.fill_range!(z_lvl_2.val, 0.0, z_lvl_qos, z_lvl_qos_stop)
-                    end
-                    z_lvl_2.val[z_lvl_qos] = y_lvl_2_val_2
-                    z_lvl.idx[z_lvl_qos] = phase_stop_2
-                    z_lvl_qos += 1
-                    y_lvl_q += 1
+            if y_lvl_i == phase_stop_2 && x_lvl_i == phase_stop_2
```
willow-ahrens, line 44: It looks like your changes made the bounds inference smarter? Is that right?

src/tensors/levels/sparselistlevels.jl
```diff
@@ -150,7 +150,11 @@ end
 summary_fiber_abbrev(lvl::VirtualSparseListLevel) = "sl($(summary_fiber_abbrev(lvl.lvl)))"
 
 function virtual_level_size(lvl::VirtualSparseListLevel, ctx)
-    ext = Extent(literal(lvl.Ti(1)), lvl.shape)
+    if lvl.Ti <: Integer
```
willow-ahrens, line 153: Could we add a TODO comment here saying that `we need to convert this to a multiple dispatch approach`?

willow-ahrens (commented):
Awesome! I think my main request is to avoid `is_continuous_extent` when you can, and instead use `measure` or `get_eps`.

PR #235, opened 2023-07-07
Replacing scansearch with Finch.scansearch

nullplay:
This pull request aims to enhance the performance of looplet levels that utilize the `scansearch` function, such as SparseList.

I found calling `scansearch` in Stepper causes runtime dispatch of `scansearch` in Julia. As a result, numerous sections of the code were impacted due to the inability to statically infer the output type of `scansearch`.

To address this issue and achieve better performance, I propose a simple solution: replacing all instances of `scansearch()` with `Finch.scansearch()`. This modification resolves the problem and results in improved overall performance.


willow-ahrens:
I am unable to find any performance difference between this and the main branch using `runjudge`. Are you perhaps running into the fact that `scansearch` is undefined in global scope when you  `eval(@finch_kernel ...)`?

I still think we should merge this, but for correctness rather than performance. Running `eval(@finch_kernel ...)` sometimes results in broken code because `scansearch`, `resize_if_smaller!`, and `fill_range!` are all internal to Finch. Could we add qualified names (`Finch.resize_if_smaller!` instead of `resize_if_smaller` or `$resize_if_smaller`) for those and double check that dead code elimination can handle function expressions of the form `module.name`?

willow-ahrens (changes requested):
I am unable to find any performance difference between this and the main branch using `benchmark/runjudge.jl`. Are you perhaps running into the fact that `scansearch` is undefined in global scope when you  `eval(@finch_kernel ...)`?

I still think we should merge this, but for correctness rather than performance. Running `eval(@finch_kernel ...)` sometimes results in broken code because `scansearch`, `resize_if_smaller!`, and `fill_range!` are all internal to Finch. Could we add qualified names (`Finch.resize_if_smaller!` instead of `resize_if_smaller` or `$resize_if_smaller`) for those and double check that dead code elimination can handle function expressions of the form `module.name`?

willow-ahrens:
I think I understand how this could lead to a performance regression without erroring: If you call `@finch_kernel` and evaluate in `Main` module, `scansearch` will be undefined. Then the compiler can't infer the output type and inference breaks. However, some kernels will never call `scansearch`, so they still run without error. However, when I call `runjudge.jl`, it only uses the `@finch` interface, and the generated code is evaluated in the `Finch` module, so `scansearch` is defined. I think my previous review is still the best fix here.

nullplay:
I observed a speedup on my custom benchmarking file that I wrote to evaluate in `Main` module. Unfortunately, I didn't see any speed improvement from `run_judge.jl` as well :-( However, as you mentioned, I believe it's still worth merging for the sake of clarity.

Regarding the dead code elimination part, I didn't fully understand its implications. I extend changes to `resize_if_smaller` and `fill_range` as well. Could you clarify if these changes are related to dead code elimination?

willow-ahrens:
I'll push a fix today, my comment was mainly about https://github.com/willow-ahrens/Finch.jl/blob/94048ac58038e3fe0571f12f0ba184b3be88a394/src/util.jl#L153

Issue #245, opened 2023-07-28
Parameterized Level Storage for Accelerators


willow-ahrens:
Parameterizing levels with an abstractarray vector type that encodes storage locations and adding a `moveto!` type.

PR #249, opened 2023-08-01
Slay the Dragon: Parallel Analysis

wraith1995:
This pull adds an analysis pass to check if parallelism is okay to use, but the properties of levels/combinator are not corect.


src/FinchNotation/FinchNotation.jl
```diff
@@ -34,10 +34,13 @@ module FinchNotation
...
+export parallelAnalysis, ParallelAnalysisResults
 
     include("nodes.jl")
     include("instances.jl")
     include("protocols.jl")
-    include("syntax.jl")
-end
\ No newline at end of file
+include("syntax.jl")
+include("analysis.jl")
```
willow-ahrens, line 45: syntax and analysis should both live in the transforms subdirectory. FinchNotation is just a module to define the AST

src/FinchNotation/analysis.jl
```diff
@@ -0,0 +1,174 @@
...
+            break
+        end
+    end
+
+    return ParallelAnalysisResults(naive, withAtomics, withAtomicsAndAssoc, tensorsNeedingAtomics, nonInjectiveAccss, nonAssocAssigns, nonConCurrentAccs)
+end
+
+
+
+#=
```
willow-ahrens, line 169: we can delete this comment now

src/tensors/levels/abstractlevel.jl
```diff
@@ -0,0 +1,18 @@
...
+#is_laminable_updater(lvl::AbstractVirtualLevel, ctx) = false
+
+
+
+is_concurrent(lvl::AbstractVirtualLevel, ctx) = true
+
+is_injective(lvl::AbstractVirtualLevel, ctx, accs) = false
+
+
+getroot(tns::AbstractVirtualLevel) = tns
```
willow-ahrens, line 15: getroot should not be defined on a level, instead define on a fiber (which would be an `AbstractVirtualArray`

willow-ahrens (changes requested):
I've requested a few changes. Could you merge master as well? Also, I think that we may want to define `AbstractVirtualArray` as a supertype of `AbstractCombinator`

willow-ahrens:
@teo, after you fix up these things I can go back and update everyone's properties.

src/FinchNotation/analysis.jl
```diff
@@ -0,0 +1,174 @@
...
+    for (root, nodeSet) in assignByRoot
+        rep = first(nodeSet)
+        tns = rep.lhs.tns
+        if rep.lhs.idxs[end] != index
+            naive = false
+            push!(tensorsNeedingAtomics, root)
+            push!(nonInjectiveAccss, rep.lhs)
+        end
+
+        # FIXME: TRACE.
```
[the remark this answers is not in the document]
wraith1995: No - that was before I learned how trace was implemented.

src/FinchNotation/FinchNotation.jl
wraith1995: There already is a syntax.jl in transforms. I'd argue analysis is not a transform, but it also doesn't matter so I will move that.

wraith1995:
@willow-ahrens:
> does this fix #243?
No. I will do that separately. 

> I've requested a few changes. Could you merge master as well? Also, I think that we may want to define `AbstractVirtualArray` as a supertype of `AbstractCombinator`

Do we really want `AbstractVirtualArray` and `VirtualAbstractArray`? I agree that we probably need something like this, but we should maybe give the type hierarchy a bit more thought. I think right now we have:

- `AbstractVirtualFiber`
- `AbstractVirtualCombinator`
- `AbstractVirtualLevel`
- `VirtualAbstractArray`

I think based on this, the logical thing is `VirtualArray` or `AbstractVirtualTensor` as the root of all of these except level, which only appears as an argument to fiber. I suppose we could do`VirtualAbstractArray` in that way.

Will `VirtualAbstractArray` be related to #245 ?

willow-ahrens:
As far as names go, let's go with `AbstractVirtualTensor` Also, I don't think we need `VirtualAbstractFiber`, as `Fiber` is not intended to be overloadable.

willow-ahrens (approved):

willow-ahrens:
@wraith1995 lmk if you're good with this PR! I still don't think I fully understand what "concurrent" means, but maybe this is a step one and we'll figure it out as we go.

src/transforms/concurrent.jl
wraith1995, line 1: Minor comment: Does is_injective belong here?

willow-ahrens: Oh, like in that file?

wraith1995: Yeah.

willow-ahrens: I think after this PR, we should do a reorganizational PR, because there's a lot of things that really should live in AbstractTensor.

wraith1995:
@willow-ahrens Looks good to me except for two minor things that we can fix later - I've added comments.

src/transforms/concurrent.jl
wraith1995: Agreed.

wraith1995: Though maybe wait on the locals stuff....

willow-ahrens: yeah I was gonna wait till you merge, but I think I can do the reorg pretty quickly.

PR #250, opened 2023-08-01
Jaeyeon/loopreduction

nullplay:
This PR implements loop reduction with two different approaches : (1) top-down and (2) bottom-up.

1. Top-down
- if Loop body does not contain loop idx
- collapse(reduce) assignment expression if lhs is declared outside of scope.

2. Bottom-up 
- if assignment itself or sequence(..., assign, ...) is the loop body,
- collapse expression if rhs is constant w.r.t loop & lhs is declared outside of scope.


src/symbolic/simplify_program.jl
```diff
@@ -111,24 +111,52 @@ function get_program_rules(alg, shash)
         (@rule sieve(true, ~a) => a),
         (@rule sieve(false, ~a) => block()), #TODO should add back skipvisitor
 
+
+        (@rule loop(~idx, ~ext::isvirtual, ~body) => begin
+            #body_contain_idx = mapreduce((node) -> node.kind === index && node == idx,
```
willow-ahrens, line 116: This comment feels like clutter

src/symbolic/simplify_program.jl
```diff
@@ -111,24 +111,52 @@ function get_program_rules(alg, shash)
         (@rule sieve(true, ~a) => a),
         (@rule sieve(false, ~a) => block()), #TODO should add back skipvisitor
 
+
+        (@rule loop(~idx, ~ext::isvirtual, ~body) => begin
+            #body_contain_idx = mapreduce((node) -> node.kind === index && node == idx, 
+            #                             (x, y) -> x || y, PostOrderDFS(body), init=false) 
+            body_contain_idx = idx ∈ getunbound(body)
+            if !body_contain_idx
```
willow-ahrens, line 119: I think these lines could be
```
if idx \not\in PostOrderDFS(body)
    ...
```

src/symbolic/simplify_program.jl
```diff
@@ -111,24 +111,52 @@ function get_program_rules(alg, shash)
...
+                decl_in_scope = Set()
+                ScopeVisitor(scope=decl_in_scope)(body)
```
willow-ahrens, line 121: I didn't realize ScopeVisitor does this. One could also do:

```
decl_in_scope = filter(!isnothing, map(node-> if @capture(node, declare(~tns, ~init)) tns end, PostOrderDFS(body)))
```

src/symbolic/simplify_program.jl
```diff
@@ -111,24 +111,52 @@ function get_program_rules(alg, shash)
...
 
+
+        (@rule loop(~idx, ~ext::isvirtual, ~body) => begin
+            #body_contain_idx = mapreduce((node) -> node.kind === index && node == idx, 
+            #                             (x, y) -> x || y, PostOrderDFS(body), init=false) 
+            body_contain_idx = idx ∈ getunbound(body)
+            if !body_contain_idx
+                decl_in_scope = Set()
+                ScopeVisitor(scope=decl_in_scope)(body) 
+                Postwalk(@rule assign(access(~lhs, ~m), +, ~b) => begin
```
willow-ahrens, line 122: This would be a place to call `is_collapsible`

src/symbolic/symbolic.jl
```diff
@@ -228,6 +228,33 @@ isinvolution(::AbstractAlgebra, ::typeof(-)) = true
...
+"""
+iscollapsable(::Any, f) = false
+iscollapsable(::AbstractAlgebra, ::typeof(or)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(and)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(|)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(&)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(+)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(*)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(min)) = true
+iscollapsable(::AbstractAlgebra, ::typeof(max)) = true
```
willow-ahrens, line 247: I think `iscollapsable` and `collapsed` should be the same function. `collapsed` should have the signature `collapsed(algebra, ::f, idx, ext, node)` and return `nothing` if not collapsible, or the collapsed form otherwise. Giving `collapsed` these arguments would allow you to write

```
collapsed(algebra, ::typeof(+), idx, ext::ContinuousExtent, node) = 
   if @capture node :call(*, ~a1..., :call(d, idx), ~a2...)
       :call(*, a1..., a2..., measure(ext))
   end
   ```

**Review by willow-ahrens — COMMENTED** [review]
Looks good! I made a comment on the signature of collapsed. Also, this is the kind of change that really needs tests before we can merge. Ideally, we would delete some of the old strength reduction rules here too (in favor of new ones)

File: src/symbolic/simplify_program.jl (hunk 2 above)
Comment (willow-ahrens) [reply src/symbolic/simplify_program.jl:119]: getunbound is fine for now though

File: src/symbolic/simplify_program.jl (hunk 3 above)
Comment (nullplay) [reply src/symbolic/simplify_program.jl]: Yeah, using ScopeVisitor is excessive for this problem. I replaced with your suggestion.

File: src/symbolic/symbolic.jl (hunk 5 above)
Comment (willow-ahrens) [reply src/symbolic/symbolic.jl]: At this point, I think if you remove iscollapsible the code will behave the same.

File: src/symbolic/simplify_program.jl [hunk 6]
```diff
@@ -111,24 +111,55 @@ function get_program_rules(alg, shash)
...
         (@rule assign(~a, ~op, cached(~b, ~c)) => assign(a, op, b)),
 
-        (@rule loop(~i, ~ext::isvirtual, assign(access(~a, ~m), $(literal(+)), ~b::isliteral)) =>
-            assign(access(a, m), +, call(*, b, measure(ext.val)))
-        ),
-        (@rule loop(~i, ~ext::isvirtual, block(~s1..., assign(access(~a, ~m), $(literal(+)), ~b::isliteral), ~s2...)) => if ortho(getroot(a), s1) && ortho(getroot(a), s2)
-            block(assign(access(a, m), +, call(*, b, measure(ext.val))), loop(i, ext, block(s1..., s2...)))
+        (@rule loop(~i, ~a::isvirtual, assign(access(~b, updater(), ~j...), ~f::isidempotent(alg), ~c)) => begin
+            if i ∉ j && i ∉ getunbound(c)
+                sieve(call(>, measure(a.val), 0), assign(access(b, updater(), j...), f, c))
```
willow-ahrens, line 161: `sieve(call(>, measure(a.val), 0), ...)` could be moved to `collapsed` as well

nullplay: What is the purpose of using a `sieve` on idempotent functions? It seems unnecessary to use a `sieve` on **non-idempotent** functions. I understand that removing the `sieve` could result in several failed tests.

Are there any differences in lowering between idempotent and non-idempotent functions?

willow-ahrens: we use sieve to avoid the empty interval.

```
max(a) = a
max(a, b) = max(a, b)
max(a, b, b) = max(a, b)
max(a, b, b, b) = max(a, b)
```

however, when we collapse +, the * operator handles the empty interval just fine:

```
a = a + 0 * b
a + b = a + 1 * b
a + b + b = a + 2 * b
```

nullplay:
I feel like this is ready to merge.

benchmark/benchmarks.jl
```diff
@@ -24,7 +24,7 @@ C = Fiber!(Dense(SparseList(Element(0.0))))
 
 @finch (C .= 0; for i=_, j=_, k=_; C[j, i] += A[k, i] * B[k, i] end)
 """
-cmd = sequence(`$(Base.julia_cmd()) --project=$(Base.active_project()) --eval $code`, stdout = IOBuffer())
+cmd = pipeline(`$(Base.julia_cmd()) --project=$(Base.active_project()) --eval $code`, stdout = IOBuffer())
```
willow-ahrens, line 27: this should be sequence

src/FinchNotation/syntax.jl
```diff
@@ -43,6 +43,8 @@ const instance_nodes = (
     dimless = :($finch_leaf_instance(dimless))
 )
 
+∂() = 1
+∂(args...) = 0
```
willow-ahrens, line 47: do we wanna provide a non-unicode alternative like `d` or `d_`?

willow-ahrens (approved):
I approve of these changes. Consider adding `d_` or `d` as an alias for the unicode d, because people sometimes find unicode hard to type.

willow-ahrens:
Jaeyeon! This PR looks great! Those rules used to be so messy and confusing, and now they are so much more robust! Thanks!

src/FinchNotation/syntax.jl
nullplay: I'll change `∂` to `d` :)

nullplay:
Solve #181

PR #252, opened 2023-08-11
Introduce Locals 

wraith1995:
Currently WIP because many minutes spent on segfaults.


wraith1995:
Not the direction we went with.

PR #254, opened 2023-08-23
add architecture tag

willow-ahrens:
Adds an architecture tag. Part 1 of n in eventually merging https://github.com/willow-ahrens/Finch.jl/pull/249


willow-ahrens:
@wraith1995 This PR is clos but not quite. It would have been nice to be more specific about concurrency by checking e.g. exclusive uses of a level, etc. One idea I had was:

only in safe mode, on declaration of a level, set a counter on the heap to zero to represent the number of iterators over that level. On unfurling that level, emit a thunk that atomically adds one to that counter and asserts the value is at most one. After we're done unfurling, the epilogue decrements the counter.

This would allow us to describe situations like when we have a level that allocates an entirely separate fiber for each index (i.e. a list-of-lists format matrix). Since each level would allocate a separate counter for it's lock, we could use the sparse list level when it's wrapped in the lists of lists format, but not if multiple threads try to access the same level. I think this style of safety check is the most correct and specific answer to whether a particular datastructure is being used correctly. It is unfortunate that this wouldn't be checked statically.

PR #256, opened 2023-08-26
WIP: Locals and copying

### LINKED ISSUE #245: Parameterized Level Storage for Accelerators

wraith1995:
fixes #245 

Todo:

Update on what is left to do:
- [x] A few tests still fail.
- [ ] In constructors, we need to use convert rather than the vector functions.
- [ ] Maybe more generally make the constructors nicer
- [x] We need to add docs, maybe.


wraith1995:
@willow-ahrens Review feedback on this is probably useful now; the main work that needs to be done to tests to pass is to adopt the broadcast/index infrastructure and to update the docs.

willow-ahrens:
what is involved in adopting the broadcast/index infrastructure?

wraith1995:
> what is involved in adopting the broadcast/index infrastructure?

When building the output data structures, we need to properly communicate the index and position types.

willow-ahrens:
hmm. So you'd like the position and index types to propagate to the output, and also the vector type? Like GPU * GPU = GPU?

willow-ahrens:
I'm envisioning something like:

`bc_output(f, a::SparseData, b::SparseData) = SparseData(f(default(a), default(b)))`

willow-ahrens:
Would you be open to my first merging a PR which makes those functions a little simpler?

wraith1995:
> Would you be open to my first merging a PR which makes those functions a little simpler?

Sure.

willow-ahrens:
Here's the PR, what do you think? https://github.com/willow-ahrens/Finch.jl/pull/269

wraith1995:
Okay. Added docs.  Running fixbot...

docs/src/fibers.md
```diff
@@ -57,11 +57,48 @@ Dense [1:4]
 ├─[4]: 0.0
 
 julia> A_fbr[:, 3]
-Dense [1:4]
-├─[1]: 4.4
-├─[2]: 0.0
-├─[3]: 5.5
-├─[4]: 0.0
+ERROR: ArgumentError: invalid index: 1.0 of type Float64
```
willow-ahrens: It looks like your implementation is mixing up value and index types here

docs/src/fibers.md
```diff
@@ -159,7 +196,33 @@ SparseCOO (0.0) [1:4,1:3]
...
 access. For random access, one should use the `SparseHash` format. A full listing
-of supported formats is described below:
+of supported formats is described after a rough description of shared common internals of level,
+relating to types and storage.
+
+## Types and Storage of Level
+
+All levels have an `indextype` and a `postype`, typically denoted as `Ti` and `Tp` in the constructors. 
+Index type is mostly self expanatory as it is ofen just `Int`, but for levels that store multiple dimensions, 
+the index type is often a subtype of a Tuple. Pos (or position) type is the type used to define the iteratiorator through any
```
willow-ahrens, line 169: spelling

docs/src/fibers.md
```diff
@@ -159,7 +196,33 @@ SparseCOO (0.0) [1:4,1:3]
...
+## Types and Storage of Level
+
+All levels have an `indextype` and a `postype`, typically denoted as `Ti` and `Tp` in the constructors. 
+Index type is mostly self expanatory as it is ofen just `Int`, but for levels that store multiple dimensions, 
+the index type is often a subtype of a Tuple. Pos (or position) type is the type used to define the iteratiorator through any 
+auxiliary data stuctures; it is named for the pos array in typical definitions of CSC or CSR.
+
+Additionally, many levels have a `VTp` or `VTi` in their constructors; these stand for vector of element type `Tp` or `Ti`. 
+More generally, levels are paramterized by the types that they use for storage. By default, all levels use `Vector`, but a user 
+could in theory all the storage types of their levels so that a fiber could be stored on a GPU or CPU or via a different
```
willow-ahrens, line 174: `could in theory all the storage types`?

src/base/broadcast.jl
```diff
@@ -93,28 +93,29 @@ broadcast_rep_child(r::SparseData) = r.lvl
 broadcast_rep_child(r::DenseData) = r.lvl
 broadcast_rep_child(r::RepeatData) = r.lvl
 
-broadcast_rep(::BroadcastRepDenseStyle, f, args) = DenseData(broadcast_rep(f, map(broadcast_rep_child, args)))
+broadcast_rep(::BroadcastRepDenseStyle, f, args) = DenseData(broadcast_rep(f, map(broadcast_rep_child, args)), Int)
```
willow-ahrens, line 96: is there ever a case where `SparseList{Int16} + SparseList{Int16} = SparseList{Int16}`, or will it always be Int? I suppose it doesn't make sense to propagate position types?

src/base/fsparse.jl
```diff
@@ -20,10 +20,10 @@ See also: [`fiber!`](@ref)
 
 ```jldoctest
 julia> println(summary(fiber(sparse([1 0; 0 1]))))
-2×2 Fiber!(Dense(SparseList(Element(0))))
+2×2 Fiber!(Dense(SparseList(Element(0, Int64))))
```
willow-ahrens, line 23: Can position type default to Int64? I'd rather not have to think about position type every time I make a level

src/base/mapreduce.jl
```diff
@@ -33,22 +33,22 @@ function reduce_rep_def(op, z, lvl::HollowData, idx, idxs...)
     end
 end
 
-reduce_rep_def(op, z, lvl::SparseData, idx::Drop, idxs...) = ExtrudeData(reduce_rep_def(op, z, lvl.lvl, idxs...))
+reduce_rep_def(op, z, lvl::SparseData, idx::Drop{Idx}, idxs...) where {Idx} = ExtrudeData(reduce_rep_def(op, z, lvl.lvl, idxs...), Idx)
```
willow-ahrens, line 36: Does Idx need to be a type parameter of Drop? I think it would be okay for it to be in the runtime here.

src/tensors/levels/repeatrlelevels.jl
```diff
@@ -92,7 +114,7 @@ end
 @inline level_ndims(::Type{<:RepeatRLELevel}) = 1
 @inline level_size(lvl::RepeatRLELevel) = (lvl.shape,)
 @inline level_axes(lvl::RepeatRLELevel) = (Base.OneTo(lvl.shape),)
-@inline level_eltype(::Type{RepeatRLELevel{D, Ti, Tp, Tv}}) where {D, Ti, Tp, Tv} = Tv
+@inline level_eltype(::Type{RepeatRLELevel{D, Ti, Tp, Tv, VTp, VTi, VTv}}) where {D, Ti, Tp, Tv, VTp, VTi, VTv} = Tv
```
willow-ahrens, line 117: Some of this parameterization is unused, but maybe we can ask aqua to fix this for us

src/tensors/levels/sparsehashlevels.jl
```diff
@@ -32,46 +34,70 @@ SparseHash (0.0) [1:3,1:3]
...
-SparseHashLevel{N, Ti, Tp, Tbl}(lvl::Lvl, args...) where {N, Ti, Tp, Tbl, Lvl} = SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, args...)
-
-SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl) where {N, Ti, Tp, Tbl, Lvl} =
-    SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, ((zero(ti) for ti in Ti.parameters)..., ))
-SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, shape) where {N, Ti, Tp, Tbl, Lvl} =
-    SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, shape, Tbl())
-SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, shape, tbl) where {N, Ti, Tp, Tbl, Lvl} =
-    SparseHashLevel{N, Ti, Tp, Tbl, Lvl}(lvl, Ti(shape), tbl, Tp[1], Pair{Tuple{Tp, Ti}, Tp}[])
+SparseHashLevel{N, Ti}(lvl, args...) where {N, Ti} = SparseHashLevel{N, Ti, postype(typeof(lvl)), Dict{Tuple{postype(typeof(lvl)), Ti}, postype(typeof(lvl))}, (memory_type(typeof(lvl))){postype(typeof(lvl)), 1}, (memory_type(typeof(lvl))){Pair{Tuple{postype(typeof(lvl)), Ti}, postype(typeof(lvl))}, 1}, typeof(lvl)}(lvl, args...)
+# FIXME: Adding pos here is not neeccesarily the right thing...
```
willow-ahrens, line 53: This is fine for now, someone should really go through and implement multilevels or something so that the hashlevel can be composed with a coolevel

src/tensors/levels/sparselistlevels.jl
```diff
@@ -94,12 +120,12 @@ function display_fiber(io::IO, mime::MIME"text/plain", fbr::SubFiber{<:SparseLis
...
 
-@inline level_ndims(::Type{<:SparseListLevel{Ti, Tp, Lvl}}) where {Ti, Tp, Lvl} = 1 + level_ndims(Lvl)
+@inline level_ndims(::Type{<:SparseListLevel{Ti, Tp,  VTp, VTi, Lvl}}) where {Ti, Tp,  VTp, VTi, Lvl} = 1 + level_ndims(Lvl)
 @inline level_size(lvl::SparseListLevel) = (level_size(lvl.lvl)..., lvl.shape)
 @inline level_axes(lvl::SparseListLevel) = (level_axes(lvl.lvl)..., Base.OneTo(lvl.shape))
-@inline level_eltype(::Type{<:SparseListLevel{Ti, Tp, Lvl}}) where {Ti, Tp, Lvl} = level_eltype(Lvl)
-@inline level_default(::Type{<:SparseListLevel{Ti, Tp, Lvl}}) where {Ti, Tp, Lvl} = level_default(Lvl)
-data_rep_level(::Type{<:SparseListLevel{Ti, Tp, Lvl}}) where {Ti, Tp, Lvl} = SparseData(data_rep_level(Lvl))
+@inline level_eltype(::Type{<:SparseListLevel{Ti, Tp,  VTp, VTi, Lvl}}) where {Ti, Tp,  VTp, VTi, Lvl} = level_eltype(Lvl)
+@inline level_default(::Type{<:SparseListLevel{Ti, Tp,  VTp, VTi, Lvl}}) where {Ti, Tp,  VTp, VTi, Lvl} = level_default(Lvl)
```
willow-ahrens, line 127: some of these functions would be much cleaner if they operated on the structs rather than the types.

src/tensors/levels/sparsecoolevels.jl
```diff
@@ -33,34 +37,73 @@ SparseCOO (0.0) [1:3,1:3]
...
+function memory_type(::Type{SparseCOOLevel{N, Ti, Tp, Tbl, VTp, Lvl}}) where {N, Ti, Tp, Tbl, VTp, Lvl}
+    return containertype(VTp)
+end
+
+function postype(::Type{SparseCOOLevel{N, Ti, Tp, Tbl, VTp, Lvl}}) where {N, Ti, Tp, Tbl, VTp, Lvl}
+    return Tp
+end
+
+function indextype(::Type{SparseCOOLevel{N, Ti, Tp, Tbl, VTp, Lvl}}) where {N, Ti, Tp, Tbl, VTp, Lvl}
+    return indextype(Ti)
```
willow-ahrens, line 94: A question about indextype. Should this perhaps just return the top index?

willow-ahrens: rather than tuple{Int}?

src/tensors/vector.jl
```diff
@@ -0,0 +1,45 @@
+containertype(::Type{T}) where T <:AbstractVector = T.name.wrapper
+
+unwrap(::Type{T}) where {T} = T
+
+
+single(::Type{Vector{S}}) where S = S[1]
```
willow-ahrens, line 6: I would much rather use standard functions here, such as `single(T) = convert(T, [1])`, rather than asking people to implement several nonstandard constructors for their new array type.

src/tensors/vector.jl
```diff
@@ -0,0 +1,45 @@
...
+empty(::Type{Vector{S}}) where S = S[]
+
+postype(::Type{Vector{S}}) where {S} = Int
```
willow-ahrens, line 9: why does vector need a postype?

src/tensors/vector.jl
```diff
@@ -0,0 +1,45 @@
...
+    if N > 1
+        return  NTuple{N, T}
+    else
+        return Tuple{T}
+    end
+end
+
+
+
+struct ForceCopyArray{T, N} <: AbstractArray{T, N}
```
willow-ahrens, line 34: Does this one deserve a separate file in `tensors/combinators`?

willow-ahrens (changes requested):

src/tensors/levels/sparsecoolevels.jl
wraith1995: `indextype` type on a tuple of types does this.

src/tensors/vector.jl
willow-ahrens: Also, if `single(T)` really is `convert(T, [1])`, why not just call `convert(T, [1])`? it's much clearer about the contents of the array

src/tensors/vector.jl
wraith1995: I think it doesn't but I was mirroring indextype for reasons unknown. Will check.

src/tensors/levels/sparsecoolevels.jl
willow-ahrens: wait, so is the argument to indextype a level or is it an index?

willow-ahrens: should we just call `first(Ti.parameters)`?

willow-ahrens: or `last(Ti.parameters)`?

src/tensors/vector.jl
wraith1995: Those should actually just be deleted - I remade them in test_moving.jl

Now, we should export something like that, but I am not sure it belongs in combinators.

docs/src/fibers.md
willow-ahrens: Also, I think the description of what `indextype` is a little unclear here. Can we narrow the scope of the indextype trait to be:

```
"""
    indextype(lvl)
Return an index type with the same flavor as those used to index the fibers contained in `lvl`.
"""
function indextype end
```

src/base/fsparse.jl
wraith1995: It does. That is actually the index type and it defaults to `Int`

wraith1995: But do you mean in the printing of the types?

src/tensors/vector.jl
wraith1995: Ah so this actually is how we determined the position type of element level. If you supply a vector that uses some other position type, then element level would have a different one and this would propagate up.

src/tensors/vector.jl
wraith1995: Because [1] creates a Vector{T}, which we should avoid. We want to directly create a value of the given vector type.  We can make `single(T) = convert(T, [1])` the default behavior though.

src/base/fsparse.jl
willow-ahrens: yeah. I think that whenever we print an Element level, we should check if the postype or indextype is equal to the default before printing them out. Most users will never use these features

src/tensors/vector.jl
wraith1995: Actually, it technically already is it turns out.

willow-ahrens: why can't we make a little Vector{T}?

docs/src/fibers.md
wraith1995: Fixed by doctest.

src/tensors/levels/sparsecoolevels.jl
wraith1995: > wait, so is the argument to indextype a level or is it an index?

Either. Indextype on a level typically ends up calling index type on a tuple of types.

wraith1995: I felt this was a clean way to deal with the variety of things that could be put there.

src/tensors/vector.jl
wraith1995: We don't want to do that on some random device because it means we leave how to do that up to Julia and lots of little Vector{T} could have consequences.

willow-ahrens: To give some context on where I'm coming from: I want someone to be able to look at the constructor for `SparseListLevel` and be able to see that `pos` is initialized to `[1]` and `idx` is initialized to `[]`. I don't think that `single` is an intuitive representation of this. If you ask ChatGPT what the `single` function does in Julia, it says:

The single function in Julia is used to convert an array of floating-point numbers to single-precision floating-point numbers. Single-precision floating-point numbers use 32 bits to represent a number, while double-precision floating-point numbers (the default in Julia) use 64 bits.

src/base/broadcast.jl
wraith1995: I am not sure.  I think it does make sense to do so (the right thing to do probably eventually ends with calling the type promotion rules), but I am not sure it will ever cause any problems.

Though on the other hand how does broadcast interact with continuous indicies (@nullplay )?

src/tensors/vector.jl
willow-ahrens: I don't think we ever call the constructor in question during moveto anyway

src/tensors/vector.jl
willow-ahrens: Ah, maybe instead we can say `postype(::ElementLevel) = eltype(memtype(ElementLevel))`?

src/base/broadcast.jl
willow-ahrens: I suppose this is also handled through size inference in the broadcast as well.

wraith1995: I think for now we should leave broadcast edge stuff until later.

willow-ahrens:
so long as you're okay with my eventually removing calls to `single` and `empty`, I will merge this.

willow-ahrens (approved):
So long as you're okay if I do some renaming and removing calls to `single` and `empty`, we can merge this

wraith1995:
> So long as you're okay if I do some renaming and removing calls to `single` and `empty`, we can merge this

Fine with me!

PR #260, opened 2023-08-29
Jaeyeon/stepjump

nullplay:
Attempt to combine Stepper(Step) and Jumper(Jump).

PR #270, opened 2023-09-07
new combine

nullplay:
Attempt to combine Stepper(Step) and Jumper(Jump).
This is a different approach compared to #260 .

Summary of changes

- Combine Step(Jump) into Stepper(Jumper) with same interface.
- Incorporate phase's lowering into stepper/jumper lowering.


nullplay:
@willow-ahrens I feel like this PR has a clearer code structure compared to previous approach! What do you think?

willow-ahrens:
This looks great! One request: Can we replace AcceptStepper and AcceptJumper with just Stepper and Jumper, and replace phase_body with stepper_body and jumper_body? I think it makes it simpler to avoid the extra wrapper looplet. We use AcceptRun to mean an output looplet for a run. I'm not sure that AcceptRun is really a good pattern, but the word Accept means different things in this PR and in AcceptRun.

test/reference64/typical_merge_gallop.txt
```diff
@@ -33,316 +33,310 @@ quote
...
     if phase_stop >= 1
         i = 1
+        if y_lvl.idx[y_lvl_q] < 1
+            y_lvl_q = Finch.scansearch(y_lvl.idx, 1, y_lvl_q, y_lvl_q_stop - 1)
+        end
+        if x_lvl.idx[x_lvl_q] < 1
+            x_lvl_q = Finch.scansearch(x_lvl.idx, 1, x_lvl_q, x_lvl_q_stop - 1)
+        end
         while i <= phase_stop
-            if y_lvl.idx[y_lvl_q] < i
```
willow-ahrens, line 37: I notice that the seek has moved. Do we think this new code is improved?

src/looplets/jumpers.jl
```diff
@@ -31,61 +52,89 @@ function lower(root::FinchNode, ctx::AbstractCompiler,  style::JumperStyle)
...
-end
-
-FinchNotation.finch_leaf(x::Jump) = virtual(x)
-
-(ctx::Stylize{<:AbstractCompiler})(node::Jump) = ctx.root.kind === loop ? JumperPhaseStyle() : DefaultStyle()
-
-function phase_range(node::Jump, ctx, ext)
-    push!(ctx.code.preamble, node.seek !== nothing ? node.seek(ctx, ext) : quote end)
-    similar_extent(ext, getstart(ext), node.stop(ctx, ext))
+function phase_range(node::AcceptJumper, ctx, ext)
```
willow-ahrens, line 112: I noticed you introduced phase_range and phase_body for AcceptJumper, but maybe we can just use Jumper_range and jumper_body?

willow-ahrens (changes requested):
Looks great! Much improved from the other approach, I agree. I had just the one comment about code style avoiding the use of AcceptJumper.

test/reference64/typical_merge_gallop.txt
nullplay: I think the new code will be faster than the previous one since it avoids calling a `scansearch` inside the while loop everytime. What I'm worried is : is lifting `scansearch` for jumper always a safe transformation?

willow-ahrens: that's not always true, the stepper might be skipped (e.g. multiply by zero)

willow-ahrens:
@nullplay this looks good, but I don't think the seek optimization is correct. Could we do the seek optimization as a separate PR and merge this without it?

PR #287, opened 2023-09-22
Fix bug with define not being in a block

### LINKED ISSUE #286: Bug with define node not in a block
Code:
~~~
A = Fiber!(Dense(Dense(SparseList(Element(0.0)))), fsprand((10, 10, 10), 0.1))
temp = 0.0
@finch begin
           for k=_, i=_, l=_
                   temp = A[i, l, k]
           end
       end
~~~

Error message: `ERROR: unimplemented (define(variable(temp), literal(0.0)))`

radha-patel:
fixes #286

### DIFF
--- a/docs/src/interactive.ipynb
+++ b/docs/src/interactive.ipynb
@@ -49,11 +49,11 @@
    "file_extension": ".jl",
    "mimetype": "application/julia",
    "name": "julia",
-   "version": "1.9.2"
+   "version": "1.6.7"
   },
   "kernelspec": {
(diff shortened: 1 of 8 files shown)


willow-ahrens (approved):
This is ready for merging as soon as tests pass! The main change here was to make the lone define a no-op, but also to address why and make sure that the lower.jl machinery can handle block statements correctly even when they are arbitrarily nested. Other syntactic transforms can canonicalize blocks to different forms as needed.

PR #294, opened 2023-09-30
Replace comparisons with wrappers

radha-patel:
Replace `<`, `<=`, `>`, `>=`, and `=` involving index comparisons with appropriate masks.


willow-ahrens:
is it possible to handle != as well, by putting a ! in front of the diagonal mask?

willow-ahrens:
it looks like some of the tests for conversions with triangle levels have failed. could you take a look at the logs to see if you can reproduce?

radha-patel:
> it looks like some of the tests for conversions with triangle levels have failed. could you take a look at the logs to see if you can reproduce?

Yep! I'm looking now

src/transforms/wrapperize.jl
```diff
@@ -42,6 +42,40 @@ function get_wrapper_rules(alg, depth, ctx)
                 end
             end
         end),
+        (@rule call(<, ~i::isindex, ~j::isindex) => begin
+            if depth(i) < depth(j)
+                access(LoTriMask(), reader(), call(-, i, 1), j)
+            elseif depth(i) > depth(j)
+                access(UpTriMask(), reader(), call(+, i, 1), j)
+            end
+        end),
```
willow-ahrens, line 51: Could we refactor each of these rules into two cases:
```
        (@rule call(<, ~i, ~j::isindex) => begin
            if depth(i) < depth(j)
                access(UpTriMask(), reader(), j, call(-, i, 1))
            end
        ),
        @rule call(<, ~i::isindex, ~j) => begin
            if depth(i) > depth(j)
                access(LoTriMask(), reader(), i, call(+, j, 1))
            end
        ),
 ```
 This code is different in a few ways:
 1. Only the inner (fast) index needs to be a vanilla index (as opposed to `i + 1`, etc.).
 2. This code flips the index order in order to always emit a column major access.

File: src/transforms/wrapperize.jl [hunk 2]
```diff
@@ -42,6 +42,40 @@ function get_wrapper_rules(alg, depth, ctx)
...
+            elseif depth(i) > depth(j)
+                access(LoTriMask(), reader(), i, j)
+            end
+        end),
+        (@rule call(==, ~i::isindex, ~j::isindex) => begin
+            access(DiagMask(), reader(), i, j)
+        end),
+        (@rule call(!=, ~i::isindex, ~j::isindex) => begin
+            call(!, access(DiagMask(), reader(), i, j))
+        end),
```
willow-ahrens, line 78: These two rules are not always emitting column major code

willow-ahrens (changes requested):
In addition to the more specific comments, none of these cases actually swap the order of i and j depending on which is bound first.

radha-patel:
@willow-ahrens I've addressed your comments!

PR #349, opened 2023-12-07
Scale, Product Array Combinator

nullplay:
Do not merge


willow-ahrens:
what's the difference between scale and product? Also, does this work for discrete extent when the factor is a positive integer?

nullplay:
1. Scale is dealing with constant coefficient and Product deals variable coefficient. Scale is similar to Offset combinator and Product is similar to Toeplitz.
2. No. We need some mechanism that restricts `scaledim` starts and ends at integer endpoint. Current `scaledim` just divide with a scale. (E.g., `scaledim(Extent(3,6), 2)` = `Extent(1.5, 3.0)`, but what we want for discrete Extent is `Extent(2, 3)`)

willow-ahrens:
Can we reverse the direction of scaledim so that it multiplies the extent, rather than dividing? My intuition is that we should be able to write `B[i] = A[i*2 + 1]`, right?

PR #362, opened 2023-12-31
Jaeyeon/newdispatch

nullplay:
Hope this fixes the continuous tests on CI.

### DIFF
--- a/src/dimensions.jl
+++ b/src/dimensions.jl
@@ -232,8 +232,8 @@ virtual_intersect(ctx, a::Dimensionless, b::Dimensionless) = b
 
 function virtual_intersect(ctx, a::Extent, b::Extent)
     Extent(
-        start = @f(max($(getstart(a)), $(getstart(b)))),
-        stop = @f(min($(getstop(a)), $(getstop(b))))
+        start = simplify(@f(max($(getstart(a)), $(getstart(b)))), ctx),
+        stop = simplify(@f(min($(getstop(a)), $(getstop(b)))), ctx)
     )
 end
 
@@ -244,8 +244,8 @@ virtual_union(ctx, a::Dimensionless, b::Dimensionless) = b
 #virtual_union(ctx, a, b) = virtual_union(ctx, promote(a, b)...)
 function virtual_union(ctx, a::Extent, b::Extent)
     Extent(
-        start = @f(min($(getstart(a)), $(getstart(b)))),
-        stop = @f(max($(getstop(a)), $(getstop(b))))
+        start = simplify(@f(min($(getstart(a)), $(getstart(b)))), ctx),
+        stop = simplify(@f(max($(getstop(a)), $(getstop(b)))), ctx)
     )
 end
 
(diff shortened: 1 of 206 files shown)


willow-ahrens:
Something weird I have found:
```
julia> hash(1.0, UInt(123))
0x5bca7c69b794f75d

julia> hash(1.0f0, UInt(123))
0x5bca7c69b794f75d

julia> hash(limit(1.0f0), UInt(123))
0xab8298daf3de8022

julia> d = Dict{Any, String}(1.0=>"hello")
Dict{Any, String} with 1 entry:
  1.0 => "hello"

julia> d[limit(1.0f0)]
"hello"

julia> for i = 1:10000
       d[i] = "bar"
       end

julia> d[limit(1.0f0)]
ERROR: KeyError: key limit(1.0) not found
Stacktrace:
 [1] getindex(h::Dict{Any, String}, key::Limit{Float32})
   @ Base ./dict.jl:482
 [2] top-level scope
   @ REPL[37]:1
```
Our problem is that `isequal(1.0, 1.0f0) == true`, which isn't really helpful for us when implementing `shash`, as we'd really like different hashes for things with different types. I'm trying this out on https://github.com/willow-ahrens/Finch.jl/pull/363.

src/util/limits.jl
```diff
@@ -166,32 +166,30 @@ Base.:(+)(x::Limit)::Limit = x
 Base.:(-)(x::Limit)::Limit = limit(-x.val, -x.sign)
 
 #Crazy julia multiple dispatch stuff don't worry about it
-limit_types = [Int8, Int16, Int32, Int64, Int128, UInt8, UInt16, UInt32, UInt64, UInt128, BigInt, Float32, Float64]
-for S in limit_types
```
willow-ahrens, line 170: This for loop avoids method ambiguities on types that define e.g. `max(a::MyType, b::Real)`. If we define `max(a::Real, b::Limit)` where `Limit <: Real`, then we get an ambiguity error. If we want to take your approach, we need to make `Limit` not be a subtype of `Number`. I understand that this is suboptimal.

willow-ahrens: Oh, I see that currently `Limit <: Number`, so `Limit` is not currently subtyping `Real`. I think this might work then. But should we have `Limit <: Real`?

src/dimensions.jl
```diff
@@ -232,8 +232,8 @@ virtual_intersect(ctx, a::Dimensionless, b::Dimensionless) = b
 
 function virtual_intersect(ctx, a::Extent, b::Extent)
     Extent(
-        start = @f(max($(getstart(a)), $(getstart(b)))),
-        stop = @f(min($(getstop(a)), $(getstop(b))))
+        start = simplify(@f(max($(getstart(a)), $(getstart(b)))), ctx),
```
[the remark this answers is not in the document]
nullplay: The CI tests on the main branch used to fail because the result of `simplify` was occasionally different on x64 Ubuntu Julia 1.6.7 when the Finch expression included a Limit. 

Therefore, I attempted to explicitly launch the `simplify` pass using this code snippet to verify if it works. In any case, it appears that your main branch now passes the continuous tests after your hash fix, so I assume the underlying issue was related to the static hash handling of the Limit type. Consequently, we can remove this change.

src/util/limits.jl
nullplay: This was also a random attempt to address the CI issues in continuous testing. While it makes the code cleaner, it introduces ambiguity. Since the CI now passes, I believe we no longer need this pull request. Let's close it.

PR #382, opened 2024-01-19
Tfc/atomic

### LINKED ISSUE #246: Atomic Array Wrapper
Add a wrapper that makes any array atomic by adding an array of locks

wraith1995:
fixes #246


willow-ahrens:
okay so in general, Lvl is the last type argument but the first struct field. Could you change SeparationLevels and AtomicLevels to match this convention? Also, could you call the array_of_atomics array something like lck for lock?

wraith1995:
@willow-ahrens Yeah no problem

willow-ahrens:
Could you also add a section in the tensor formats docs that describes what these levels do? Perhaps call them "modifier levels".

rayegun:
Sorry to barge in here but will this level allow for the fine grained saxpy that we have in SuiteSparse:GraphBLAS? It uses atomics for fine grained (within vector) parallelism

wraith1995:
@rayegun Maybe? sSend me a code snippet and I can tell you.

rayegun:
@wraith1995 The code base is laid out really funky and complex (lots of C style templating) but it is described succinctly around page 15 of https://github.com/DrTimothyAldenDavis/GraphBLAS/blob/stable/Doc/toms_parallel_grb2.pdf. The fine grained tasks for saxpy.

wraith1995:
@rayegun I think this might violate the lifecycle constraints as written, but we might be able to come up with something that didn't and used atomics to get parallelism. I would have to think about it more closely.

wraith1995:
@rayegun If you could write this in Finch without parallelism, I can see if atomics would let you parallelize this, but I am not sure how to even write that in Finch.

wraith1995:
@rayegun I've thought about this more. With the addition of one of two other levels, we can probably do a version of this.

.github/workflows/CI.yml
```diff
@@ -48,6 +48,11 @@ jobs:
             ${{ runner.os }}-test-
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
+      - name: Set and Retrieve Github ENV variables
+        shell: bash
+        run: |
+          export JULIA_NUM_THREADS=2
```
willow-ahrens, line 54: This line doesn't do anything because only the GITHUB_ENV is persistent across steps

docs/src/guides/tensor_formats.md
```diff
@@ -99,6 +101,7 @@ The "Level Format Name" is the name of the level datatype. Other columns have de
 | ✅     | Indicates the level is ready for serious use. |
 | ⚙️     | Indicates the level is experimental and under development. |
 | 🕸️     | Indicates the level is deprecated, and may be removed in a future release. |
+| ❓   | Indicates a feature of a level only works in certain circumstances. |
```
willow-ahrens, line 104: I think we could just use a checkbox for all of those, right?

src/tensors/levels/atomiclevel.jl
```diff
@@ -0,0 +1,250 @@
...
+function assemble_level!(lvl::VirtualAtomicLevel, ctx, pos_start, pos_stop)
+    pos_start = cache!(ctx, :pos_start, simplify(pos_start, ctx))
+    pos_stop = cache!(ctx, :pos_stop, simplify(pos_stop, ctx))
+    idx = freshen(ctx.code, :idx)
+    lockVal = freshen(ctx.code, :lock)
+    push!(ctx.code.preamble, quote 
+              Finch.resize_if_smaller!($(lvl.locks), $(ctx(pos_stop))) 
+              @inbounds for $idx = $(ctx(pos_start)):$(ctx(pos_stop))
+                lockVal = make_lock(eltype($(lvl.AVal)))
+                if !isnothing(lockVal)
```
willow-ahrens, line 138: can we skip this isnothing and just set `make_lock` default to a pass-through? e.g. `make_lock(Bool) = false`

src/tensors/levels/atomiclevel.jl
```diff
@@ -0,0 +1,250 @@
...
+
+function instantiate(fbr::VirtualSubFiber{VirtualAtomicLevel}, ctx, mode::Updater, protos)
+    (lvl, pos) = (fbr.lvl, fbr.pos)
+    sym = freshen(ctx.code, lvl.ex, :after_atomic_lvl)
+    atomicData = freshen(ctx.code, lvl.ex, :atomicArraysAcc)
+    lockVal = freshen(ctx.code, lvl.ex, :lockVal) 
+    dev = lower(virtual_get_device(ctx.code.task), ctx, DefaultStyle())
+    return Thunk(
+        preamble = quote  
+            $atomicData =  promote_val_to_lock($dev, $(lvl.locks), $(ctx(pos)), eltype($(lvl.AVal)))
```
willow-ahrens, line 219: could we rename `promote_val_to_lock` to `get_lock` and `get_lock!` to `aquire_lock!`?

willow-ahrens (dismissed):
LGTM, I have only a few minor aesthetic requests. I haven't seen this `promote_val_to_lock` pattern before but it really fits here. Thanks!

docs/src/guides/tensor_formats.md
willow-ahrens: since they do technically support these if their contained level does.

src/tensors/levels/atomiclevel.jl
willow-ahrens: `make_lock(Lock) = Lock()`

src/tensors/levels/atomiclevel.jl
```diff
@@ -0,0 +1,254 @@
...
+    pos_start = cache!(ctx, :pos_start, simplify(pos_start, ctx))
+    pos_stop = cache!(ctx, :pos_stop, simplify(pos_stop, ctx))
+    idx = freshen(ctx.code, :idx)
+    lockVal = freshen(ctx.code, :lock)
+    push!(ctx.code.preamble, quote 
+              Finch.resize_if_smaller!($(lvl.locks), $(ctx(pos_stop))) 
+              @inbounds for $idx = $(ctx(pos_start)):$(ctx(pos_stop))
+                lockVal = make_lock(eltype($(lvl.AVal)))
+                if lockVal == false
+                    break
```
willow-ahrens, line 139: why do we break?

PR #390, opened 2024-01-29
Swizzle back in `copyto!`

### LINKED ISSUE #387: Question regarding 3d swizzle array
Hi @willow-ahrens,

I've got a question about permuting 3d swizzle array. Here's a short code where permuting a dense array differs from a Finch array, but resulting shapes should be equal (for other combinations of dimensions it worked as expected):

```julia
using Finch

A = zeros(2, 4, 3)
A[1,:,:] = [0.0 0.0 4.4; 1.1 0.0 0.0; 0.0 0.0 0.0; 3.3 0.0 0.0]
A[2,:,:] = [1.0 0.0 0.0; 0.0 0.0 0.0; 0.0 1.0 0.0; 3.3 0.0 0.0]

permutation = (3, 1, 2)

new_shape_1 = size(permutedims(A, permutation))


t = Tensor(Dense(SparseList(SparseList(Element(0.0)))), A)
st = swizzle(t, permutation...)
# materialize swizzle
new_shape_2 = size(Tensor(Dense(SparseList(SparseList(Element(0.0)))), st))

print(new_shape_1, new_shape_2)  # they don't equal
```

Is my reasoning correct?

mtsokol:
Hi @willow-ahrens,

I took a look at #387 and I think that `return ret.body` causes omitting "swizzling" back to the desired dimensionality in some `copyto!` functions. I just started learning how it works so I might have misunderstood it.

fixes #387

### DIFF
--- a/src/interface/copy.jl
+++ b/src/interface/copy.jl
@@ -28,13 +28,11 @@ function permutedims(src::Tensor, perm)
 end
 
 function Base.copyto!(dst::Union{Tensor, AbstractArray}, src::SwizzleArray{dims}) where {dims}
-    ret = copyto!(swizzle(dst, invperm(dims)...), src.body)
-    return ret
+    return copyto!(swizzle(dst, invperm(dims)...), src.body)
 end
 
 function Base.copyto!(dst::SwizzleArray{dims1}, src::SwizzleArray{dims2}) where {dims1, dims2}
-    ret = copyto!(swizzle(dst, invperm(dims2)[collect(dims1)]...), src.body)
-    return ret
+    return copyto!(swizzle(dst, invperm(dims2)[collect(dims1)]...), src.body)
 end
 
 function Base.copyto!(dst::SwizzleArray{dims}, src::Union{Tensor, AbstractArray}) where {dims}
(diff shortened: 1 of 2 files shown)


willow-ahrens:
Awesome, thanks so much for the fix! Could you add a quick test case and then I'll merge?

willow-ahrens (changes requested):
Looks great! Just needs at least one test

mtsokol:
@willow-ahrens I added a test that checks `permutedims` for 3d tensor with different shapes. I think that the current fix isn't sufficient as some of the test cases fail with incorrect shapes compared to `Base.permutedims`.

willow-ahrens:
I think the main issue here is that Finch seems to drop the swizzle wrapper on the output. This is a little bit of a tricky fix, I'll need to look at it tonight and get back to you.

mtsokol:
Thank you! For me it's a bit difficult to follow code considering that ’tocopy!’ both returns a value and mutates passed argument for the same purpose. And it seems that IT matters which one is used in different places.

willow-ahrens:
The main issue is here: https://github.com/willow-ahrens/Finch.jl/blob/0a536c7f43ef76ffd55ca0a927e654466e1efda8/src/transforms/wrapperize.jl#L181-L188 

In these lines of code, Finch is stripping the swizzle off of the tensor, and doesn't have a way to re-wrap the tensor before return.

PR #391, opened 2024-01-29
Gpu draft

hochawa:
Just share the current state.


gdalle:
@willow-ahrens is there still interest in GPU support here?

PR #394, opened 2024-01-30
Add a return statement to finch

willow-ahrens:
To help make it so copyto! doesn't drop swizzles, we need return statements in Finch so that we can have a place to add the swizzle back on.

https://github.com/willow-ahrens/Finch.jl/issues/387


willow-ahrens:
Before I make this work across all of finch, what does everyone think about this change? I have created a return type for Finch, so that we can be specific about which tensors get returned from Finch, and also so that if we need to strip wrappers from arrays we can still return the correct arrays in the end (as is the case with SwizzledArray and eventually maybe a BlockedArray or SymmetricArray type)

willow-ahrens:
A note: to really make this work, I think we technically need to fuse the 'trim' functionality together with declare whenever it is semantically meaningful

wraith1995:
@willow-ahrens At a first glance this looks good to me, but we should test/document multiple returns.

willow-ahrens:
There are still many bugs to track down, but I sorta wanted to get a feel what people thought, because it's a big change

wraith1995:
@willow-ahrens Also, how is this interacting with lifecycles? I think return should only be callable on tensors that are in a certain state/are not being iterated over.

willow-ahrens:
i did need to modify lifecycles to get it to work

willow-ahrens:
i mostly declared the return as a read access to the tensor

wraith1995:
@willow-ahrens I see. It feels fine to me on a frontend level - like this is an unobjectionable feature though returning nothing should still be an option! That is also unclear just skimming around.

On an implementation level, what should distinguished returned tensors from ones that we modify but don't return? Should anything? Or should this be represented in the lifecycles via making all returned ones read whereas non-retuned ones are left in whatever state they are last put in?

willow-ahrens:
One interesting thing that has shown up in this PR: for better or worse we've been passing scalars by value, not reference. Not returning the scalar means the new immutable value it contains will not be passed to the user. Also, unreturned tensors aren't getting trimmed. A sorta tangential response to your question

willow-ahrens:
but I think that the only difference between returned and unreturned tensors should be whether the user cares about accessing the new immutable values, such as dimensions

willow-ahrens:
@wraith1995 Currently this PR requests a return explicitly. I'm working on a way to insert a return into `@finch` and `@finch_code` automatically with any tensors declared in scope. `@finch_kernel` and any explicit calls to `execute` would still need an explicit return. lmk what you think, I'm merging this to fix the copyto swizzle bug

Issue #381, opened 2024-01-18
Off By One Array


willow-ahrens:
As an alternative to CIndex (described here: https://willowahrens.io/Finch.jl/dev/guides/interoperability/) let's create a wrapper vector that subtracts one.

willow-ahrens:
Ultimately, I think what we want is a barebones special case of https://github.com/JuliaArrays/MappedArrays.jl/blob/master/src/MappedArrays.jl

willow-ahrens:
closed by #405

PR #405, opened 2024-02-07
API: Introduce `OffByOneVector`

mtsokol:
Addresses: https://github.com/willow-ahrens/Finch.jl/issues/381

Hi @willow-ahrens @hameerabbasi,

This PR adds the `OffByOneVector` struct that allows to use NumPy arrays as indices arrays without a copy.

### DIFF
--- a/src/tensors/vectors.jl
+++ b/src/tensors/vectors.jl
@@ -1,0 +1,36 @@
+using Base: @propagate_inbounds
+
+struct OffByOneVector{T} <: AbstractVector{T}
+    data::AbstractVector{T}
+end
+
+@propagate_inbounds function Base.getindex(vector::OffByOneVector{T},
+                                           index::Int) where {T}
+    return vector.data[index] + 0x01
+end
+
+@propagate_inbounds function Base.getindex(vector::OffByOneVector{T},
+                                           index::Vararg{Int}) where {T}
+    return vector.data[index...] + 0x01
+end
+
+@propagate_inbounds function Base.setindex!(vector::OffByOneVector{T},
+                                            val::T,
+                                            index::Int) where {T}
+    vector.data[index] = val - 0x01
+end
+
+@propagate_inbounds function Base.setindex!(vector::OffByOneVector{T},
+                                            val::T,
+                                            index::Vararg{Int}) where {T}
+    vector.data[index...] = val - 0x01
+end
+
+Base.parent(vector::OffByOneVector{T}) where {T} = vector.data
+Base.size(vector::OffByOneVector{T}) where {T} = size(vector.data)
+Base.axes(vector::OffByOneVector{T}) where {T} = axes(vector.data)
+
+function moveto(vector::OffByOneVector{T}, device) where {T}
+    data = moveto(vector.data, device)
+    return OffByOneVector{T}(data)
+end
--- a/test/test_issues.jl
+++ b/test/test_issues.jl
@@ -176,9 +176,9 @@ using CIndices
     let 
         m = 4; n = 3; ptr_c = [0, 3, 3, 5]; idx_c = [1, 2, 3, 0, 2]; val_c = [1.1, 2.2, 3.3, 4.4, 5.5];
 
-        ptr_jl = unsafe_wrap(Array, reinterpret(Ptr{CIndex{Int}}, pointer(ptr_c)), length(ptr_c); own = false)
-        idx_jl = unsafe_wrap(Array, reinterpret(Ptr{CIndex{Int}}, pointer(idx_c)), length(idx_c); own = false)
-        A = Tensor(Dense(SparseList{CIndex{Int}}(Element{0.0, Float64, CIndex{Int}}(val_c), m, ptr_jl, idx_jl), n))
+        ptr_jl = OffByOneVector(ptr_c)
+        idx_jl = OffByOneVector(idx_c)
+        A = Tensor(Dense(SparseList{Int}(Element{0.0, Float64, Int}(val_c), m, ptr_jl, idx_jl), n))
 
         @test A == [0.0 0.0 4.4; 1.1 0.0 0.0; 2.2 0.0 5.5; 3.3 0.0 0.0]
     end
--- a/docs/src/guides/interoperability.md
+++ b/docs/src/guides/interoperability.md
@@ -16,11 +16,51 @@ Julia](https://docs.julialang.org/en/v1/manual/embedding/#Working-with-Arrays).
 
 However, for sparse array formats, it's not just a matter of subtracting one
 from the index, as the internal lists of indices, positions, etc all start from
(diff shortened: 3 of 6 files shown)


src/tensors/vectors.jl
```diff
@@ -0,0 +1,23 @@
+struct OffByOneVector{T} <: AbstractVector{T}
+    data::AbstractVector{T}
+end
+
+function Base.getindex(vector::OffByOneVector{T}, index::Int) where {T}
+    return vector.data[index] + 1
+end
+
+function Base.getindex(vector::OffByOneVector{T}, index::Vararg{Int}) where {T}
+    return vector.data[index...] + 1
```
willow-ahrens, line 10: add and subtract 0x01, UInt8 won't widen smaller ints with promotion.

src/tensors/vectors.jl
```diff
@@ -0,0 +1,23 @@
+struct OffByOneVector{T} <: AbstractVector{T}
+    data::AbstractVector{T}
+end
+
+function Base.getindex(vector::OffByOneVector{T}, index::Int) where {T}
```
willow-ahrens, line 5: all getters should use `Base.@propagate_inbounds`

src/tensors/vectors.jl
```diff
@@ -0,0 +1,23 @@
...
+end
+
+function Base.setindex!(::OffByOneVector{T}, ::Any, ::Int) where {T}
+    throw(ErrorException("`OffByOneVector` is a read only object."))
+end
+
+function Base.setindex!(::OffByOneVector{T}, ::Any, ::Vararg{Int}) where {T}
+    throw(ErrorException("`OffByOneVector` is a read only object."))
+end
+
```
willow-ahrens, line 20: let's also implement `moveto` for this vector, it's documented in this Finch somewhere.

src/tensors/vectors.jl
```diff
@@ -0,0 +1,23 @@
...
+function Base.getindex(vector::OffByOneVector{T}, index::Int) where {T}
+    return vector.data[index] + 1
+end
+
+function Base.getindex(vector::OffByOneVector{T}, index::Vararg{Int}) where {T}
+    return vector.data[index...] + 1
+end
+
+function Base.setindex!(::OffByOneVector{T}, ::Any, ::Int) where {T}
```
willow-ahrens, line 13: let's make these mutable so we can write to a python-readable object and convert back if we need to

src/tensors/vectors.jl
```diff
@@ -0,0 +1,23 @@
+struct OffByOneVector{T} <: AbstractVector{T}
```
willow-ahrens, line 1: Could we add some docs about this to the manual instead of talking about CIndex?

willow-ahrens (commented):
This is a great start! I think the big asks from me are to allow setindex and update the manual section about CIndex. We should be able to delete CIndex from this package after this.

rayegun:
Quick question: What were the downsides you faced with CIndex? In my mind both solutions have slightly different usecases, but I'm curious to hear your thoughts @willow-ahrens

src/tensors/vectors.jl
mtsokol: Done!

src/tensors/vectors.jl
mtsokol: Done!

src/tensors/vectors.jl
mtsokol: Done!

src/tensors/vectors.jl
mtsokol: Done!

src/tensors/vectors.jl
willow-ahrens: I don't think any code that relies on CIndices will break if we remove it from Finch, since CIndices is an external package. There's some commented code that users it in binsparse.jl which could be replaced with this vector, and a test that uses them where we could also add the solutions developed here. I think we should also simplify the docs to avoid the unsafe pointer stuff.

willow-ahrens: We could leave the docs, but add a warning that the approach is no longer recommended

willow-ahrens:
CIndex is a great approach when we can't wrap the Vector type (e.g. interacting with SparseMatrixCSC). However, using it in such a situation requires unsafe pointer stuff. To reinterpret with CIndex safely, we need a reinterpreted array. Also, Cindex propagates through the Finch code, so it's unclear if it would be a performant kernel. On the other hand, OffByOneArray only adds or subtracts 1 when we read and write, which is an operation we have already minimized.

src/tensors/vectors.jl
willow-ahrens: We would really only use it when we need to interact with julia code that doesn't let us modify the vector type. We might consider deferring to the docs CIndex already has, or moving our docs to CIndex.jl

mtsokol: Sure! I removed `CIndex` from the test and `binsparse.jl`, and added a note to `CIndex` docs.

src/tensors/vectors.jl
mtsokol: Did you mean `... + 1 + 0x01 - 0x01`?

willow-ahrens: a similar approach is `vector.data[index...] + T(1)`

src/tensors/vectors.jl
willow-ahrens: https://documenter.juliadocs.org/stable/showcase/#Admonitions

willow-ahrens: I would use a warning admonition

test/test_issues.jl
```diff
@@ -176,9 +176,9 @@ using CIndices
     let 
         m = 4; n = 3; ptr_c = [0, 3, 3, 5]; idx_c = [1, 2, 3, 0, 2]; val_c = [1.1, 2.2, 3.3, 4.4, 5.5];
 
-        ptr_jl = unsafe_wrap(Array, reinterpret(Ptr{CIndex{Int}}, pointer(ptr_c)), length(ptr_c); own = false)
-        idx_jl = unsafe_wrap(Array, reinterpret(Ptr{CIndex{Int}}, pointer(idx_c)), length(idx_c); own = false)
-        A = Tensor(Dense(SparseList{CIndex{Int}}(Element{0.0, Float64, CIndex{Int}}(val_c), m, ptr_jl, idx_jl), n))
+        ptr_jl = OffByOneVector(ptr_c)
```
willow-ahrens, line 179: Nice! So much more straightforward!

willow-ahrens (changes requested):
Looks great! A few minor changes and then we can merge

src/tensors/vectors.jl
mtsokol: Done! I added `!!! warning` admonition.

willow-ahrens (approved):

Issue #134, opened 2023-03-07
Common Subexpression Elimination on input IR


willow-ahrens:
In the new IR, it is easier to express operations that reuse reads to a tensor multiple times. For example, we can write symmetric matvec on a triangle as:

```
@finch begin
   @loop i j begin
      y[i] += A[i, j] * x[j]
      y[j] += A[i, j] * x[i]
   end
end
```

it would be great to automatically transform to the much more efficient form:

```
@finch begin
   a.= 0
   a[] = A[i, j]
   @loop i j begin
      y[i] += a[] * x[j]
      y[j] += a[] * x[i]
   end
end
```

PR #408, opened 2024-02-09
Rrpatel cse

willow-ahrens:
fixes https://github.com/willow-ahrens/Finch.jl/issues/134


willow-ahrens:
Cool, this looks great!

PR #409, opened 2024-02-09
Finer-grained Parallel analysis

wraith1995:


willow-ahrens:
What's the status here? Should we try to merge soon?

willow-ahrens:
Note: I think that some fixes to separationlevel were merged to main that might not be in this PR.

wraith1995:
@willow-ahrens There are two issues that I want to fix here and both are parallelism problems.

1. Atomics still has one edge case - I am afraid that element level and its ilk might need independent atomic options or else more complex machinery than I thought to deal with scalars.
2. Separate level is not actually concurrent at the moment. I am isolating the issue, but as a starting point I have a function that is fine when the number of threads is 1 but fails at 2 on HB/bcspwr07. 
```
function spgemm_finch_gustavson_kernel_parallel(A, B)
    # @assert Threads.nthreads() >= 2
    z = default(A) * default(B) + false
    C = Tensor(Dense(Separate(SparseList(Element(z)))))
    w = Tensor(Dense(Element(z)))
    @finch_code begin
        C .= 0
        for j=parallel(_)
            w .= 0
            for k=_, i=_; w[i] += A[i, k] * B[k, j] end
            for i=_; C[i, j] = w[i] end
        end
    end
    @finch begin
        C .= 0
        for j=parallel(_)
            w .= 0
            for k=_, i=_; w[i] += A[i, k] * B[k, j] end
            for i=_; C[i, j] = w[i] end
        end
    end
    return C
end

```

wraith1995:
Update: It looks like the above code fails even when the output array is dense dense so this really is just an issue of parallelism I suspect.

willow-ahrens:
Haha i just got the email notification for that commit message, and I think my whole life for the past two years could probably be summarized as "many tiny fixes to compile"

wraith1995 (commented):
This looks good to me.

PR #415, opened 2024-02-14
API: Add `eltype` for `SwizzleArray`

mtsokol:
Hi @willow-ahrens,

Just a nitpick one-line change that I noticed in `Tensor` class in `finch-tensor`.


willow-ahrens (approved):
LGTM!

Issue #228, opened 2023-06-12
Add docs for typical use cases


willow-ahrens:
i.e. how to construct a fiber, how to convert between fibers.

willow-ahrens:
Some ideas for documentation structure attached
[structure1.md](https://github.com/willow-ahrens/Finch.jl/files/13895195/structure1.md)
[structure2.md](https://github.com/willow-ahrens/Finch.jl/files/13895196/structure2.md)
[structure3.md](https://github.com/willow-ahrens/Finch.jl/files/13895197/structure3.md)
[structure4.md](https://github.com/willow-ahrens/Finch.jl/files/13895198/structure4.md)
[structure5.md](https://github.com/willow-ahrens/Finch.jl/files/13895199/structure5.md)

PR #426, opened 2024-02-20
Create a conversion function from fiber to vector-of-tuples

kylebd99:
Per the title.

### DIFF
--- a/src/tensors/fibers.jl
+++ b/src/tensors/fibers.jl
@@ -120,7 +120,7 @@ postype(fbr::AbstractVirtualFiber) = postype(fbr.lvl)
 allocator(fbr::AbstractVirtualFiber) = allocator(fbr.lvl)
 
 struct LabelledTree
-    key 
+    key
     node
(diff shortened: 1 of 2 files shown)


kylebd99 (commented):
Not sure where a test for this function might live. Let me know if there's a preferred spot for that.

willow-ahrens:
Thanks for the PR! Summarizing our discussion today, let's convert this into documentation PR for some of the functions in https://github.com/willow-ahrens/Finch.jl/blob/main/src/interface/fsparse.jl, as well as countstored, nnz, dropdefaults, and perhaps some of the sparseArray conversions.

willow-ahrens:
@kylebd99 I'm closing this PR as we already have the ffindnz functionality, and I'll add an issue for the documentation request. Thanks for your help with this!

willow-ahrens:
The issue is here: https://github.com/willow-ahrens/Finch.jl/issues/228

Issue #427, opened 2024-02-21
`MethodError` in 1-D Tensor indexing


mtsokol:
Hi @willow-ahrens,

I noticed that 1-D Tensor indexing fails with `MethodError`, here's a code to reproduce:
```julia
using Finch

a = range(1,100,step=1) |> collect
b = Finch.Tensor(Finch.Dense(Finch.Element(0)), a)

a[30]
b[30]

a[[1,2,3]]
b[[1,2,3]] # fails with ERROR: MethodError: no method matching keys(::IndexLinear, ::Tensor{DenseLevel{Int64, ElementLevel{0, Int64, Int64, Vector{Int64}}}})

a[range(start=10, stop=20)]
b[range(start=10, stop=20)] # fails with ERROR: MethodError: no method matching keys(::IndexLinear, ::Tensor{DenseLevel{Int64, ElementLevel{0, Int64, Int64, Vector{Int64}}}})
```

Is it something easy to fix?

---

I also noticed that for n-dimensional tensors all dimensions must be provided when indexing (for example, `t1` which is 2-D must be indexed with `t1[i1, i2]`, where for `t1[i1]` it fails with `AssertionError: ndims(arr) == length(inds)`). I don't mind keeping it that way as I can infer `:` slices for the implicit dimensions on the Python side.

willow-ahrens:
# fails with ERROR: MethodError: no method matching keys(::IndexLinear, ::Tensor{DenseLevel{Int64, ElementLevel{0, Int64, Int64, Vector{Int64}}}})

^ I think this means we're missing a fallback for indexing with vectors. 

We want for `getindex(x, i::Vector)` to hit the same fallback as `getindex(A, i::Vector, j::Int)`, but instead it's hitting
the same fallback that we would hit with `getindex(A, i::Vector{Int})` or `getindex(A, i::Vector{CartesianIndex})`, or `getindex(A, i::Array{CartesianIndex})`. All of those methods attempt to index a multidimensional array with the linear index, so we could conceivably do something like:
```
julia> A = [11 12 13; 14 15 16; 17 18 19]
3×3 Matrix{Int64}:
 11  12  13
 14  15  16
 17  18  19

julia> A[[1, 2, 3, 4]]
4-element Vector{Int64}:
 11
 14
 17
 12
 ```
 
[...]

PR #429, opened 2024-02-22
API: Add list indexing for 1D Tensors

mtsokol:
Addressing #427

Hi @willow-ahrens,

This PR adds specialized `getindex` and `to_indices` for `Tensor` and `SwizzleArray`, to enable list indexing for 1D objects.
There's no need to check whether `Tensor`/`SwizzleArray` is 1D here, because indexing +2D `Tensor` with a list/vector results in `AssertionError: ndims(arr) == length(inds)`.


willow-ahrens:
LGTM! @mtsokol I'll push a release too

PR #436, opened 2024-02-24
Merge Identical Runs in SparseRLE

willow-ahrens:
Cool


willow-ahrens:
@nullplay what do you think of this? The approach here is to merge the runs at the end. I'd like to merge this PR and then reorganize the test suite a bit, but in the future we could also add a level flag to skip deduplication.

PR #437, opened 2024-02-26
Fix SparseVBLLevel bug

radha-patel:


willow-ahrens (approved):
LGTM! Thanks for the fix!

PR #438, opened 2024-02-27
API: Add `fill_value` and `fill_value_type` config argument to `Base.similar`

mtsokol:
Hi @willow-ahrens,

This PR adds `fill_value` and `fill_value_type` configuration argument to `Base.similar` to match https://docs.julialang.org/en/v1/base/arrays/#Base.similar.

Right now the change of data type can be done with:
```julia
tns = ... # let's assume it's Int64
new_tns = copyto!(similar(tns, Finch.FillValueConfig(0, Float64)), tns)
```

Let me know if a tuple is better than a dedicated struct, but I think it's more descriptive.


willow-ahrens:
Also, not in this pr of course, but what do you think of changing the term `default` in finch to be `fill_value`?

mtsokol:
> Also, not in this pr of course, but what do you think of changing the term `default` in finch to be `fill_value`?

I like this idea! `fill_value` is more precise in my opinion.

mtsokol:
> I think the tuple is not a good fit because `similar(zeros(5, 5), (5, 5))` is a valid way to specify sizes. Let's add a `fill_value` kwarg, I think that is the best approach here.

Do you mean keyword only argument `fill_value` that accepts a tuple or a dedicated struct as in this PR?
```julia
similar(A, (3, 3), fill_value=FillValueConfig(0, Int32))
similar(A, (3, 3), fill_value=(0, Int32))
```

willow-ahrens:
`similar(A, Int32, (3, 3), fill_value=0)`

willow-ahrens:
Should we just support `similar(A, dtype, ...)` and `similar(A, (fill_value, dype), ..)` and nothing else?

willow-ahrens:
Open to opinions on that. I think the Config struct is too high-powered, so I'm happy with either the `fill-value` kwarg or the two styles with the tuple I mentioned above.

willow-ahrens (changes requested):
Looks good to me, but similar should preserve size if size is unspecified. One question: should similar(Tensor(SparseList(Element(0.0), PlusOneArray(ptr), PlusOneArray(idx))) preserve the PlusOneArrays? Currently it wont

src/tensors/levels/atomiclevels.jl
```diff
@@ -25,7 +25,8 @@ AtomicLevel(lvl::Lvl) where {Lvl} = AtomicLevel{Vector{Base.Threads.SpinLock}, L
 # AtomicLevel{AVal, Lvl}(atomics::AVal, lvl::Lvl) where {Lvl, AVal} =  AtomicLevel{AVal, Lvl}(lvl, atomics)
 Base.summary(::AtomicLevel{AVal, Lvl}) where {Lvl, AVal} = "AtomicLevel($(AVal), $(Lvl))"
 
-similar_level(lvl::Atomic{AVal, Lvl}) where {Lvl, AVal} = AtomicLevel{AVal, Lvl}(similar_level(lvl.lvl))
+similar_level(lvl::Atomic{AVal, Lvl}, fill_value, eltype::Type) where {Lvl, AVal} =
```
willow-ahrens, line 28: There should be a version of this one that passes dims through

src/tensors/levels/denselevels.jl
```diff
@@ -34,8 +34,10 @@ DenseLevel{Ti}(lvl::Lvl, shape) where {Ti, Lvl} = DenseLevel{Ti, Lvl}(lvl, shape
 const Dense = DenseLevel
 
 Base.summary(lvl::Dense) = "Dense($(summary(lvl.lvl)))"
-similar_level(lvl::DenseLevel) = Dense(similar_level(lvl.lvl))
-similar_level(lvl::DenseLevel, dims...) = Dense(similar_level(lvl.lvl, dims[1:end-1]...), dims[end])
+
+similar_level(lvl::DenseLevel, fill_value, eltype::Type) = Dense(similar_level(lvl.lvl, fill_value, eltype))
```
willow-ahrens, line 38: This one should use the pre-existing dimension in lvl.

willow-ahrens: alternately, you could delete these versions of similar_level and only use the one that takes dims

willow-ahrens (commented):
A few pointers

src/tensors/levels/denselevels.jl
mtsokol: Ok, I tried to remove `similar_level` functions without `dims` argument, but I'm not sure if we can do this. There's one place here:
https://github.com/willow-ahrens/Finch.jl/blob/11cecd0641d2d408463be3ec3617aa232732c8a3/src/tensors/levels/separatelevels.jl#L131
where I don't know what `dims` I should pass. How can I know the dimensionality of `lvl`?
I tried:
```julia
$sym = similar_level($(lvl.ex).lvl, $(virtual_level_default(lvl)), $(virtual_level_eltype(lvl)), $(virtual_level_size(lvl, ctx)))

# and different variations of the last argument:
$(virtual_level_size(lvl, ctx)...)
$(virtual_level_size(lvl, ctx))...
```
But none of them compile/expand macro without an error. Maybe I could treat the remaining levels as zero size with `[0]` repeated `level_ndims(lvl)` times?

willow-ahrens: Oh no! Yes, okay, in that situation you could pass in `level_size`

willow-ahrens: actually, the appropriate thing to do would be (I know this is arcane)

```
$sym = similar_level(
    $(lvl.ex).lvl,
    $(ctx(virtual_level_default(lvl))),
    $(virtual_level_eltype(lvl)),
    $(map(ctx, map(getstop, virtual_level_size(lvl, ctx)))...)
)
```

willow-ahrens: Since you have the level already constructed you could also do:
```
$sym = similar_level(
    $(lvl.ex).lvl,
    level_default($(lvl.ex).lvl),
    level_eltype($(lvl.ex).lvl),
    level_size($(lvl.ex).lvl)
)
```

src/tensors/levels/atomiclevels.jl
mtsokol: Done!

src/tensors/levels/denselevels.jl
mtsokol: Thank you!
The final version that worked is:
```
$sym = similar_level(
    $(lvl.ex).lvl,
    level_default(typeof($(lvl.ex).lvl)),
    level_eltype(typeof($(lvl.ex).lvl)),
    level_size($(lvl.ex).lvl)...
)
```

willow-ahrens: I think thats the cleanest in that situation

willow-ahrens:
You're going to have a little trouble merging because I changed the constructor tests to be more generic. However, this should help ensure that the constructor tests cover all of the cases of similar. Could you take the tests for similar that you have and put them into the new constructor test format? They could just go in the inner loop as another case.

mtsokol:
> One question: should similar(Tensor(SparseList(Element(0.0), PlusOneArray(ptr), PlusOneArray(idx))) preserve the PlusOneArrays? Currently it wont

Hmm, in my opinion it shouldn't.

> Could you take the tests for similar that you have and put them into the new constructor test format?

Sure! Let me rebase and update the test suite.

mtsokol:
@willow-ahrens I updated the test suite with `similar` test cases. A few comments:
- Exhaustive `similar` API testing is placed in `basic_levels` section (where I check every combination of arguments)
- For the remaining sections only one `similar` call is performed to test if a given format has the correct `similar_level` implementation
- I removed type constraints in `similar_level` implementation for `AtomicLevel` and `SeparateLevel` levels. For example, the `Lvl` in `Separate{Lvl, Val}` enforces one specific level definition, where it can actually change downstream in `similar_level` calls (e.g. when `eltype` argument was passed, nested `Element{...}` will change). Let me know if this type change can be annotated in a different way.
- I think there's a bug in `SparseBand` that causes `copyto!` fail for that format: https://github.com/willow-ahrens/Finch.jl/issues/443
- I excluded `SingleList` and `SingleRLE` as calling `similar` with these levels throws an error `"SingleListLevels can only be updated once"`

test/test_constructors.jl
```diff
@@ -62,6 +62,42 @@
                 @test Structure(fbr) == Structure(Tensor(Lvl{Int16}(Element(0.0))))
                 @test Structure(fbr) == Structure(Tensor(Lvl(Element(0.0), Int16(0))))
                 @test Structure(fbr) == Structure(Tensor(Lvl{Int16}(Element(0.0), 0)))
+
+                if key == "SingleList" || key == "SingleRLE"
+                    continue  # don't test similar for Single*
+                end
+                if key == "SparseBand"
```
willow-ahrens, line 69: This one should be fixed shortly so we can remove this continue

test/test_constructors.jl
```diff
@@ -62,6 +62,42 @@
                 @test Structure(fbr) == Structure(Tensor(Lvl{Int16}(Element(0.0))))
                 @test Structure(fbr) == Structure(Tensor(Lvl(Element(0.0), Int16(0))))
                 @test Structure(fbr) == Structure(Tensor(Lvl{Int16}(Element(0.0), 0)))
+
+                if key == "SingleList" || key == "SingleRLE"
```
willow-ahrens, line 66: These are failing because you have changed the fill so you are trying to write a bunch of non-fill elements (all zero) to the level, but singlelist only supports one non-fill element.

willow-ahrens: Perhaps let's just test everything but the copyto call for these and call it a a day haha. we can just move these a few blocks down.

willow-ahrens (changes requested):
all good with me, only minor changes requested.

src/tensors/fibers.jl
```diff
@@ -326,7 +326,11 @@ end
 Base.summary(fbr::Tensor) = "$(join(size(fbr), "×")) Tensor($(summary(fbr.lvl)))"
 Base.summary(fbr::SubFiber) = "$(join(size(fbr), "×")) SubFiber($(summary(fbr.lvl)))"
 
-Base.similar(fbr::AbstractFiber) = Tensor(similar_level(fbr.lvl))
-Base.similar(fbr::AbstractFiber, dims::Tuple) = Tensor(similar_level(fbr.lvl, dims...))
+Base.similar(fbr::AbstractFiber) = similar(fbr, default(fbr), eltype(fbr), size(fbr))
+Base.similar(fbr::AbstractFiber, eltype::Type) = similar(fbr, default(fbr), eltype, size(fbr))
```
willow-ahrens, line 330: either we should attempt to `convert(eltype, default(fbr))` here or we should use `zero(eltype)`.

src/tensors/fibers.jl
```diff
@@ -326,7 +326,11 @@ end
...
+Base.similar(fbr::AbstractFiber, fill_value, eltype::Type) = similar(fbr, fill_value, eltype, size(fbr))
+Base.similar(fbr::AbstractFiber, dims::Tuple) = similar(fbr, default(fbr), eltype(fbr), dims)
+Base.similar(fbr::AbstractFiber, eltype::Type, dims::Tuple) = similar(fbr, default(fbr), eltype, dims)
```
willow-ahrens, line 333: same here

willow-ahrens (changes requested):
another small tweak

willow-ahrens:
Okay, I made a few small requests but this should be good to merge after those. This is waiting on fixes to sparsebandlevel

src/tensors/fibers.jl
mtsokol: I think `convert(eltype, default(fbr))` is a better option. Done!

src/tensors/fibers.jl
mtsokol: Done!

test/test_constructors.jl
mtsokol: Done!

test/test_constructors.jl
mtsokol: I wanted to move the `if` statement a few blocks below but I noticed that `SingleList` complains when placed in a >=2-dim tensor: https://github.com/willow-ahrens/Finch.jl/issues/447

mtsokol:
> Okay, I made a few small requests but this should be good to merge after those. This is waiting on fixes to sparsebandlevel

Hmm, after rebasing I noticed that two `conversions` tests for `SparseList{Separate}` fail. Both of them fail with:
```
ERROR: propagate_copies reached unrecognized expression level_size(tmp_lvl_2.lvl)...
```
I guess it's coming from:
```
$sym = similar_level(
    $(lvl.ex).lvl,
    level_default(typeof($(lvl.ex).lvl)),
    level_eltype(typeof($(lvl.ex).lvl)),
    level_size($(lvl.ex).lvl)...
)
```
that we implemented for `Separate` level. Do you know how it could be fixed? (Whether to change `similar_level()` call here, or modify `propagate_copies`, or change all `similar_level` functions from vararg to tuple for `dims` argument)

willow-ahrens:
I think we should write the call without the splat, I think it's better to keep the generated code with a more barebones syntax. If you can get the version that calls e.g. virtual_level_size to work, that version makes code with no splats.

test/test_constructors.jl
mtsokol: Thank you! Now it works as expected.

mtsokol:
@willow-ahrens It worked with `$(map(ctx, map(getstop, virtual_level_size(lvl, ctx)))...)`, I also updated `test/reference*/` files. The PR is ready from my side!

willow-ahrens (approved):

PR #442, opened 2024-02-27
Small fix to fsparse_impl

kylebd99:
In the previous implementation, there was a small bug which resulted in incorrectly sorted index vectors. This was because the reverse that happens on line 35 isn't re-reversed latter on.


willow-ahrens:
Awesome, could we add one small test that demonstrates the fix (so I don't break it again) to `test_issues.jl`?

kylebd99:
Of course, just added the test case that was giving me trouble!

PR #454, opened 2024-03-08
Add an einsum function

willow-ahrens:
Draft einsum proposal. Note, the current implementation does not support in-place operations because Finch High-level AST does not support in-place operations.


willow-ahrens:
@kylebd99 it looks like this uncovered a few bugs, but otherwise what do you think of my current approach?

src/interface/lazy.jl
```diff
@@ -147,9 +150,50 @@ function tensordot(A::LazyTensor{T1, N1}, B::LazyTensor{T2, N2}, idxs; mult_op=*
...
+    end
+    out_fields_in_input_order = [s_to_idx[s_idx] for s_idx in s_idxs if s_idx in out_idxs]
+    t_prod = mapjoin(immediate(mult_op), factors...)
+    t_sum = aggregate(immediate(add_op), immediate(init), t_prod, reduce_fields...)
+    t_ordered = reorder(relabel(t_sum, out_fields_in_input_order...), out_fields...)
+    S = fixpoint_type(add_op, init, t_ordered)
+    return LazyTensor{S}(identify(t_ordered), Tuple(extrude), init)
+end
+=#
+
```
willow-ahrens, line 196: @kylebd99 I'm ready to merge this, but I'm not sure what to do with this function-based interface. Would you like to keep an interface like this? Should we implement it with the functions in einsum.jl?

willow-ahrens (commented):
All set, just not sure what to do with the function-based interface.

src/interface/lazy.jl
kylebd99: I think we can ditch this interface for the moment. Going forward, we'll do optimization on the HL language anyways, and it should be easy to identify an einsum-like query in that language if we need to optimize at that level for some reason. From the python interface perspective, any potential einsum frontend will presumably just immediately translate into an  aggregate(mapjoin(...)) query in HL.

PR #456, opened 2024-03-08
Wma/better fusion

willow-ahrens:


kylebd99:
I would support merging this branch. Currently, an expression like compute(sum(.*(A[:,:,nothing], B[:, nothing,:], C[nothing,:,:]))) will first materialize the internal multiplication before computing the aggregate which wreaks havoc on performance.

If we're worried about the aggressiveness, I could see about writing a version which only pulls up mapjoins "one level".

willow-ahrens:
no, let's just go for it, full fusion, full steam ahead!
*we'll fix stuff later

Issue #464, opened 2024-03-13
Implement Tensor Dot


kylebd99:
As a precursor to the einsum issue, implementing tensor dot will make operations like mat mul more straightforward to express.

We'll be following the array interface described [here](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.tensordot.html).

PR #465, opened 2024-03-13
Kbd-tensor-dot

kylebd99:
Per issue #464, this PR implements the tensordot function from the array interface. 

To do this, I had to make some changes to the format inference code which I need a second pair of eyes on.  Specifically, lines 189-223 of compute.jl. 

The previous version of mapjoin format inference didn't take the indices of the arguments into account. For instance, in the matrix mult kernel, there's an interior mapjoin:
```
C[i,j,k] = A[i,j] * B[k,j]
```
The format inferred for C should have three levels, but it previously only had a two levels because it assumed both concordized and equal index sets (i.e. it assumed element-wise operations).


src/interface/compute.jl
```diff
@@ -187,7 +187,41 @@ function (ctx::SuitableRep)(ex)
...
+                output_def = DenseData(output_def)
+            elseif typeof(style) == Finch.MapRepSparseStyle
+                output_def = SparseData(output_def)
+            elseif typeof(style) == Finch.MapRepRepeatStyle
+                output_def = RepeatData(output_def)
+            elseif typeof(style) == Finch.MapRepExtrudeStyle
+                output_def = ExtrudeData(output_def)
+            end
+        end
+        return output_def
```
willow-ahrens, line 224: I think this would probably be more cleanly represented with extrude_rep, but I need to investigate a little more here. Does your test case exemplify the problem you faced?

src/interface/eager.jl
```diff
@@ -51,6 +51,26 @@ function Base.reduce(op::Function, bc::Broadcasted{FinchStyle{N}}; dims=:, init
...
 end
 
+
+function tensordot(A::Tensor, B::Tensor, idxs; kw...)
+    bc_A = broadcasted(identity, A)
+    bc_B = broadcasted(identity, B)
+    tensordot(bc_A, bc_B, idxs; kw...)
+end
+
+function tensordot(bc_A::Broadcasted{FinchStyle{N1}}, bc_B::Broadcasted{FinchStyle{N2}}, idxs; mult_op=*, add_op=+, init = initial_value(mult_op, Float64)) where {N1, N2}
```
willow-ahrens, line 61: Because users can now use a lazy fusion interface, maybe we don't need to support e.g. reduce(broadcast) or tensordot(broadcast), especially if it's only going to fall back to the lazy fusion interface.

src/interface/eager.jl
```diff
@@ -51,6 +51,26 @@ function Base.reduce(op::Function, bc::Broadcasted{FinchStyle{N}}; dims=:, init
     end
 end
 
+
+function tensordot(A::Tensor, B::Tensor, idxs; kw...)
+    bc_A = broadcasted(identity, A)
```
willow-ahrens, line 56: Can we instead do `compute(tensordot(lazy(A), lazy(B), idxs, kw...))`?

src/interface/lazy.jl
```diff
@@ -99,6 +99,36 @@ function Base.reduce(op, arg::LazyTensor{T, N}; dims=:, init = initial_value(op,
...
+    reduce_fields = []
+    for i in eachindex(A_idxs)
+        B_fields[B_idxs[i]] = A_fields[A_idxs[i]]
+        push!(reduce_fields, A_fields[A_idxs[i]])
+    end
+    AB = mapjoin(immediate(mult_op), relabel(A.data, A_fields), relabel(B.data, B_fields))
+    AB_reduce = aggregate(immediate(add_op), immediate(init), AB, reduce_fields...)
+    S = fixpoint_type(add_op, init, AB_reduce)
+    return LazyTensor{S}(identify(AB_reduce), extrude)
+end
```
willow-ahrens, line 130: nice!

willow-ahrens (changes requested):
This looks good! I think the major thing is to double check whether the fixes to the format computation are reimplementing existing functionality.

src/interface/compute.jl
kylebd99: Yeah, the test case should show this problem. I think new functionality is needed because the format selection is dependent on the named indices, i.e. in order to differentiate between A[i,j]*B[j,k] and A[i,j]*B[i,j]. However, the way I've written it here doesn't necessarily fit with the broader code style. I could put this logic in traits.jl potentially. I'd just need to include index information in an updated map_rep interface.

src/interface/eager.jl
kylebd99: Makes sense to me, that feels cleaner. I'll remove this version.

src/interface/eager.jl
kylebd99: Done.

willow-ahrens (approved):
Thanks Kyle!

PR #470, opened 2024-03-15
API: Add `permutedims(::SwizzleArray)`

mtsokol:
Hi @willow-ahrens,

With Finch.jl `0.6.17` I managed to run first element-wise ops (`.*`) with broadcasting in a `@compile` lazy mode! Here I wanted to clarify the eager mode.

AFAIR The current approach is:
- If a `Finch.jl` function receives (at least one) `LazyTensor` input, then the output is also `LazyTensor` (lazy mode)
- If all inputs are `SwizzleArray`/`Tensor` then it's an eager mode. `Finch.jl` will wrap it in `lazy()`, call desired op, and call `compute` to get `Tensor` output.

For the latter I got some:
```
MethodError: no method matching permutedims(::SwizzleArray...)
```

I see that `eager.jl` operates on `Tensor` input but in `finch-tensor` we made a decision that everything is a `SwizzleArray`.

Should I make `eager.jl` also accept `SwizzleArray` inputs? (Check `permutedims` in this PR) Or would you like to approach it differently?

---

P.S. I added `ndims(::LazyTensor{T, N})` as calling `jl.ndims(obj)` is more convenient than `jl.ndims(jl.typeof(obj))` IMO

Also, I started wondering, how to implement `size(::Finch.LazyTensor{T, N})`.

### DIFF
--- a/src/interface/lazy.jl
+++ b/src/interface/lazy.jl
@@ -11,7 +11,14 @@ end
 LazyTensor{T}(data, extrude::NTuple{N, Bool}) where {T, N} = LazyTensor{T, N}(data, extrude)
 
 Base.ndims(::Type{LazyTensor{T, N}}) where {T, N} = N
+Base.ndims(tns::LazyTensor) = ndims(typeof(tns))
 Base.eltype(::Type{<:LazyTensor{T}}) where {T} = T
+Base.eltype(tns::LazyTensor) = eltype(typeof(tns))
+
+Base.size(::LazyTensor) =
+    throw(ErrorException("Base.size is not supported for LazyTensor. Call `compute()` first."))
+
+Base.getindex(::LazyTensor, i...) = throw(ErrorException("Lazy indexing is not supported. Call `compute()` first."))
 
 function identify(data)
     lhs = alias(gensym(:A))
@@ -38,6 +45,8 @@ function LazyTensor{T}(arr::Tensor) where {T}
     LazyTensor{eltype(arr), ndims(arr)}(tns, extrude)
 end
 LazyTensor(data::LazyTensor) = data
+
+swizzle(arr::LazyTensor, dims...) = permutedims(arr, dims)
 
 Base.sum(arr::LazyTensor; kwargs...) = reduce(+, arr; kwargs...)
 Base.prod(arr::LazyTensor; kwargs...) = reduce(*, arr; kwargs...)
@@ -197,6 +206,7 @@ function Base.permutedims(arg::LazyTensor{T, N}, perm) where {T, N}
     idxs = [field(gensym(:i)) for _ in 1:N]
     return LazyTensor{T, N}(reorder(relabel(arg.data, idxs...), idxs[perm]...), arg.extrude[perm])
 end
+Base.permutedims(arr::SwizzleArray, perm) = swizzle(arr, perm...)
 
 Base.:+(
     x::LazyTensor,


willow-ahrens:
> Should I make eager.jl also accept SwizzleArray inputs?

In short, yes. However, not all eager operations should be routed through the lazy(compute) interface. The lazy interface has a high overhead. For example, I think get/setindex has a more direct implementation that finch is currently using that we should stick with, because those ops are usually inexpensive and the overhead of scheduling them is probably way higher than just doing the op.

The following functions have direct implementations
- getindex/setindex because they are inexpensive
- copyto! and dropdefaults! because they are fundamental operations that are called by `Tensor` constructors
- isequal and == because I want a stable source of truth for these two

permutedims feels like it is similar to copyto!, and I'd like to keep it's direct eager implementation. However, `SwizzleArray` is itself also a form of laziness. Therefore, I propose the following functionalities:

1. `permutedims(LazyTensor)` -> LazyTensor
2. `swizzle(LazyTensor)` -> LazyTensor
3. `lazy(arr::Swizzle{<:Tensor})` -> permutedims(lazy(arr.body), arr.dims)
4. `permutedims(Tensor, dims)` -> eager permutation
5. `permutedims(arr::Swizzle, dims)` -> swizzle(arr, dims...)

Does this feel reasonable? I think we need to adjust the current codebase to handle 2 and 5.

mtsokol:
Works for me! 

One question: Does it mean that `getindex/setindex` only have eager versions? Can these operations be placed inside `@compile` decorated function?

src/interface/lazy.jl
```diff
@@ -11,7 +11,9 @@ end
 LazyTensor{T}(data, extrude::NTuple{N, Bool}) where {T, N} = LazyTensor{T, N}(data, extrude)
 
 Base.ndims(::Type{LazyTensor{T, N}}) where {T, N} = N
+Base.ndims(::LazyTensor{T, N}) where {T, N} = N
```
willow-ahrens, line 14: This is usually handled by the `Base.AbstractArray` class, but sometimes making things subtype `AbstractArray` is more trouble than it's worth. The fallback is also sometimes written as `Base.ndims(tns::LazyTensor) = ndims(typeof(tns))`

I'm fine with this code as is just thought I'd elaborate on what's going on. You can read more about this here:

https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array

src/interface/lazy.jl
```diff
@@ -189,6 +191,8 @@ function Base.copyto!(dst::AbstractArray, src::LazyTensor{T, N}) where {T, N}
     return LazyTensor{T, N}(reformat(immediate(dst), src.data), src.extrude)
 end
 
+Base.permutedims(swizzle_arr::SwizzleArray{dims, <:Tensor}, perm) where {dims} =
+    compute(permutedims(permutedims(LazyTensor(swizzle_arr.body), dims), perm))
```
willow-ahrens, line 195: These two lines can just be

`Base.permutedims(arr::SwizzleArray, perm) = swizzle(arr, perm...)`

willow-ahrens (changes requested):
Only one small change requested! Looks good!

willow-ahrens:
Oh, in my review, I forgot that we should also add `swizzle(arr::LazyTensor, dims...) = permute_dims(arr, dims)`

willow-ahrens:
> Also, I started wondering, how to implement size(::Finch.LazyTensor{T, N}).

> One question: Does it mean that getindex/setindex only have eager versions? Can these operations be placed inside @compile decorated function?

LazyTensors currently only track `extrude`, i.e. whether the size is equal to one. I do think they could also track `size`, but I'm not sure yet whether it's a good idea. I would leave this alone for now because I think it's clearer to only keep track of the information that we are using to decide what computation is requested. However, I agree that it certainly would be very possible to track the size of the tensors along with their extruded dimensions. I just think that size(LazyTensor) should error out for now with a helpful error message that says to call `compute` on the lazy tensor.

In the future, we should add windowing operations to our lazy language to support `getindex` and `setindex`. Currently, this is not implemented. One way to implement these would be to understand them as multiplication by a [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix) or a similar selection mask with special properties that the compiler can analyze. let's make an issue for this and ask @kylebd99 as well for advice.

kylebd99:
An indexing operator definitely makes sense to add at some point. It matches the last operator from relational algebra that's missing in our high level language, the selection operator. In the long run, it'll be important to have this because pushdown of selection operations would be pretty fundamental for getting good performance when people do use index slicing.

willow-ahrens:
would it make sense to capture indexing with permutation matrices? I ask because that's a more minimal change

willow-ahrens:
i.e. 
```
x[3:5] = [0 0 1 0 0 0;
          0 0 0 1 0 0;
          0 0 0 0 1 0] * x
```

kylebd99:
In my opinion, this approach obfuscates the relative simplicity of the operation and might make it harder to reason about performance. But, it is more in keeping with the style of finch at the moment, so I'm not sure.

willow-ahrens:
Yeah, it could make things more complicated than they are worth. When we get to the point that indexing needs to become lazy, maybe it's worth sketching out what different notations would look like for transformation, optimization and analysis of the query notation. For the next 3 weeks I think that eager index  is fine though!

src/interface/lazy.jl
mtsokol: Sure! I added a fallback-style function.

src/interface/lazy.jl
mtsokol: Ah, right - done!

mtsokol:
@willow-ahrens I added missing `swizzle` and `permutedims` implementations from [your list](https://github.com/willow-ahrens/Finch.jl/pull/470#issuecomment-1999892679).

I also added `getindex` and `size` with more verbose error messages.

willow-ahrens (approved):
yep, this looks great! Thanks!

PR #473, opened 2024-03-18
Fix `SwizzleArray` broadcasting

mtsokol:
Hi @willow-ahrens,

In the eager mode, I think there's an issue with `broadcast(fn, ::SwizzleArray, ::SwizzleArray)`. It complained about missing functions, but with them in place it fails when performing `copyto!`.

The `test/test_issues.jl` change reproduces the issue.

### DIFF
--- a/src/interface/copy.jl
+++ b/src/interface/copy.jl
@@ -31,6 +31,16 @@ function Base.permutedims(src::Tensor, perm)
     copyto!(dst, swizzle(src, perm...))
 end
 
+# I'm not sure if this is the right way for specializing types (breaking Union from below) but
+# without it there's:
+# MethodError: copyto!(::Vector{Float64}, ::Finch.SwizzleArray{(2, 3, 1), Tensor{DenseLevel{Int64,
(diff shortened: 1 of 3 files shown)


willow-ahrens:
The issue here is more that the swizzle wasn't declaring a FinchStyle broadcast style. I added a definition for that.

In general, most of the "interface" overloads for `Tensor` (i.e. getindex, setindex, map, broadcast, reduce, copy), should also be defined for `SwizzleArray`, `PermissiveArray`, `OffsetArray`, etc. I think the right way forward on that is to add an abstract class like `CompiledArray` that everything can inherit from for the purposes of defining interfaces in terms of the Finch compiler.

The fixes I've added so far should be good for now though!

willow-ahrens:
Also, I've chosen not to define `length` on Finch tensors because most of the tensors require cartesian indexing to work properly, and length tends to be used for linear indexing. If at some point we want to do a linear indexing + length PR we could add them though!

mtsokol:
Thank you! In this version `broadcast` with `SwizzleArrays` works for me! 

I think the error comes from `copyto!(A, A_sw)` line.

willow-ahrens:
Oops! Yeah I forgot how julia `Array` cannot be resized

PR #479, opened 2024-03-20
Kbd-index-extension

kylebd99:
This PR adds a minimal form of lazy indexing to allow for empty dimensions.


willow-ahrens:
Looks good to me! Could you also add an eager version for Tensor and SwizzleArray? I think the specialization is something like `Base.getindex(tns::Tensor, idxs::Vararg{<:Union{Colon, Nothing}}) = compute(lazy(tns)[idxs...])`

kylebd99:
What's the intended behavior for eager? It seems like the :/nothing construct is primarily meant for expressing lazy computation.

willow-ahrens:
I think it would just eagerly insert a length-1 dimension? I was asking mostly because all of the constructs usually have an eager counterpart. Though no worries if you'd rather not add it here, we can skip it for now and add it later.

willow-ahrens (approved):

PR #489, opened 2024-04-08
Wma/fix 488

### LINKED ISSUE #488: should we support fsprand(n, m, p) when n * m > typemax(Int)?
Hi @willow-ahrens,

I'm working on benchmarks and for `1_000_000x1_000_000` I'm not able to move from COO format (that is returned by `fsprand`) to `CSC`:

```julia
using Finch

M = Tensor(Dense(SparseList(Element(0.0))), fsprand(1_000_000, 1_000_000, 1e-05));
```
where the last line runs infinitely.

willow-ahrens:
fixes #488

### DIFF
--- a/src/interface/fsparse.jl
+++ b/src/interface/fsparse.jl
@@ -191,9 +191,9 @@ function fsprand_erdos_renyi_gilbert(r::AbstractRNG, T, M::Tuple, p::AbstractFlo
     n = prod(M, init=1.0)
     q = 1 - p
     #We wish to sample nnz from binomial(n, p).
-    if n <= typemax(Int64)*(1 - eps())
+    if n <= typemax(Int)*(1 - eps())
         #Ideally, n is representable as an Int
-        _n = prod(M, init=Int64(1))
+        _n = Int(prod(M))
         nnz = rand(r, Binomial(_n, p))
     else
         #Otherwise we approximate
--- a/src/interface/fsparse.jl
+++ b/src/interface/fsparse.jl
@@ -80,7 +80,7 @@ Create a random sparse tensor of size `m` in COO format. There are two cases:
     also `p`).
     - If `p` is an integer, exactly `p` nonzeros are distributed uniformly at
     random throughout the tensor (and hence the density of nonzeros is exactly
-    `p / prod(m)`).
+    `p / prod(M)`).
 Nonzero values are sampled from the distribution specified by `rfn` and have the
 type `type`. The uniform distribution is used in case `rfn` is not specified.
 The optional `rng` argument specifies a random number generator.
@@ -126,26 +126,13 @@ function fsprand_erdos_renyi_sample_knuth(r::AbstractRNG, M::Tuple, nnz::Int)
     I = ntuple(n -> Vector{typeof(M[n])}(undef, nnz), N)
 
     k = 1
-    function init(n, i...)
-        if n == 0
-            for m = 1:N
-                I[m][k] = i[m]
-            end
-            k += 1
-        else
-            for i_n = 1:M[n]
-                init(n - 1, i_n, i...)
-                if k > nnz
-                    break
-                end
-            end
-        end
-    end
-    init(N)
-
     function sample(n, p::Float64, i...)
         if n == 0
-            if rand(r) * p < k 
+            if k <= nnz
+                for m = 1:N
+                    I[m][k] = i[m]
+                end
+            elseif rand(r) * p < k 
                 l = rand(r, 1:nnz)
                 for m = 1:N
                     I[m][l] = i[m]
@@ -156,9 +143,6 @@ function fsprand_erdos_renyi_sample_knuth(r::AbstractRNG, M::Tuple, nnz::Int)
             m = M[n]
             for i_n = 1:m
                 sample(n - 1, ((p - 1) * m) + i_n, i_n, i...)
-                if k > nnz
-                    break
-                end
             end
         end
     end
@@ -179,7 +163,7 @@ function fsprand_erdos_renyi_sample_self_avoid(r::AbstractRNG, M::Tuple, nnz::In
         i = ntuple(n -> rand(r, 1:M[n]), N)
         push!(S, i)       
         if length(S) > k
-            k = length(S)
+            k += 1
             for m = 1:N
                 I[m][k] = i[m]
             end
--- a/src/interface/fsparse.jl
+++ b/src/interface/fsparse.jl
@@ -109,67 +109,129 @@ fsprand_parse_type(r, T::Type, args...) = fsprand_parse_shape(r, (T,), (), args.
 fsprand_parse_type(r, args...) = fsprand_parse_shape(r, (), (), args...)
 
 fsprand_parse_shape(r, T, M, m, args...) = fsprand_parse_shape(r, T, (M..., m), args...)
-fsprand_parse_shape(r, T, M, p::AbstractFloat, rfn=rand) = fsprand_impl(r, T, M, p, rfn)
-fsprand_parse_shape(r, T, M) = throw(ArgumentError("No float p given to fsprand"))
+fsprand_parse_shape(r, T, M, p::AbstractFloat) = fsprand_parse_shape(r, T, M, p, rand)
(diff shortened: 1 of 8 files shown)


willow-ahrens:
It looks like fsprand is erroring due to integer overflow on 32bit systems with this test case.

willow-ahrens:
this is due to randsubseq, not Finch. We need to implement our own randsubseq or improve the existing implementation

mtsokol:
It looks that it's strictly due to 32bit integer being a default one on Windows.

It seems that `length(A)` gets an overflow and returns `-727379968`. But then isn't it a bug in Julia?

```julia
length(CartesianIndices((1000000, 1000000)))
```
Gives `-727379968` of 32bit integer. But I think it should anticipate which size to use and return 64bit integer.

willow-ahrens:
the main challenge is that all of these algorithms assume the entire collection we are sampling from is representable in memory.

mtsokol:
I added custom implementation to make sure index variables are `Int64` for now. Weirdly it fails with:
```
BoundsError: attempt to access 1000000×1000000 CartesianIndices{2, Tuple{Base.OneTo{Int32}, Base.OneTo{Int32}}} at index [38056]
```
which doesn't really make any sense, considering that 38056 fits in Int32.

mtsokol:
[I received a feedback on Julia Discourse.](https://discourse.julialang.org/t/boundserror-in-cartesianindices-on-32-bit-windows/112777/4?u=expine) The issue here is that internally it still checks the index against the length, so it looks like it's going to be on 32-bit anyway.

mtsokol:
One option is that we can build index ourselves, in the line where error occurs:
```julia
push!(S, A[i += ceil(Int64, s)])
```
do instead something like:
```julia
i += ceil(Int64, s)
i_tuple = Tuple{N}
for idx, dim in enumerate(dimensions):
    i_tuple[idx] = i % dim
    i = i // dim
push!(S, A[i_tuple...])
```
This way we can index A with a tuple instead of linearized index that can exceed 32bit.

willow-ahrens:
We may want to move towards a custom implementation for tuples of ints

willow-ahrens:
Im also happy to take a crack at this, if you like

mtsokol:
Let me take care of this! It's a small tweak to the existing implementation.

src/interface/fsparse.jl
```diff
@@ -130,6 +130,38 @@ end
...
+
+randsubseq(r::AbstractRNG, A::AbstractArray{T}, p::Real) where {T} =
+    randsubseq!(r, T[], A, p)
+
+# Fill S (resized as needed) with a random subsequence of A, where
+# each element of A is included in S with independent probability p.
+function randsubseq!(r::AbstractRNG, S::AbstractArray, A::AbstractArray, p::Real)
+    Base.require_one_based_indexing(S, A)
+    0 <= p <= 1 || throw(ArgumentError("probability $p not in [0,1]"))
+    n = prod(size(A), init=Int64(1))  # length
```
willow-ahrens, line 143: This line will overflow (int64 is not enough in the case we want a sparse matrix with int64 indices)

src/interface/fsparse.jl
```diff
@@ -130,6 +130,38 @@ end
...
+    p == 1 && return copyto!(resize!(S, n), A)
```
willow-ahrens, line 144: this line is fine to use prod(size(A)) because we know it won't overflow because we are about to materialize it.

willow-ahrens:
I'm concerned we need to use a different algorithm. We can't just call `randexp` and convert to an integer, I'm not even sure that's correct for Int64. I don't think we can expect it to perform well after the float mantissa loses precision, around $n=2^53$ this algorithm is only going to select $i$ equal to multiples of 2, 4, 8, etc.

mtsokol:
FYI The original implementation does:
```julia
s = randexp(r) * L
s >= n - i && return S
push!(S, A[i += ceil(Int,s)])
```

willow-ahrens:
Yes, don't think the original algorithm is correct for sparse arrays with large dimension. The problem is that the function `f(x) = ceil(Int, x::Float)` is not actually capable of producing all possible integers.

willow-ahrens:
we might want to use something like https://github.com/JuliaStats/StatsBase.jl/blob/60fb5cd400c31d75efd5cdb7e4edd5088d4b1229/src/sampling.jl#L256C1-L278C44, or we may also want to consider . This would allow us to customize the simple algorithm for tuples of ints, with correct behavior even for `fsprand(typemax(Int64), typemax(Int64), (10.0/typemax(Int64))/typemax(Int64))`. 

The above algorithm is for sampling k numbers without replacement. However, we still need to accurately generate the number of numbers we will need. We need a binomial distribution, but we need to be careful with how we call that too, since we want to sample from way more than n.

However, it's debatable whether we should do any of this, because even though we expect the number of output nonzeros to be addressable, we aren't prepared for the case where the number of elements overflows.

We may want to instead introduce a variant of `fsprand` that allows us to specify the exact number of nonzeros, then use the sampling $k$ without replacement algorithm I linked here.

willow-ahrens:
Here's my issue:

```
julia> using Finch
Precompiling Finch
  1 dependency successfully precompiled in 24 seconds. 13 already precompiled.

julia> t = 1000
1000

julia> countstored(fsprand(t, t, (100/t)/t))
108

julia> countstored(fsprand(t, t, (100/t)/t))
108

julia> countstored(fsprand(t, t, (100/t)/t))
99

julia> t = typemax(Int64)
9223372036854775807

julia> countstored(fsprand(t, t, (100/t)/t))
0

julia> countstored(fsprand(t, t, (100/t)/t))
0

julia> countstored(fsprand(t, t, (100/t)/t))
0
```

hameerabbasi:
We have an implementation here if that helps:
https://github.com/pydata/sparse/blob/0b58d4b023d299750c324f8250ab2cdb661aeea9/sparse/pydata_backend/_utils.py#L223

willow-ahrens:
@hameerabbasi I don't think the algorithm you mentioned can handle this either:

```
willow@Willows-Mac-mini ~ % poetry show 
joblib        1.4.0  Lightweight pipelining with Python functions
llvmlite      0.42.0 lightweight wrapper around basic LLVM functionality
numba         0.59.1 compiling Python code using LLVM
numpy         1.26.4 Fundamental package for array computing in Python
scikit-learn  1.4.2  A set of python modules for machine learning and data mining
scipy         1.13.0 Fundamental algorithms for scientific computing in Python
sparse        0.15.1 Sparse n-dimensional arrays for the PyData ecosystem
threadpoolctl 3.4.0  threadpoolctl
willow@Willows-Mac-mini ~ % poetry run python3
Python 3.11.8 (main, Feb  6 2024, 21:21:21) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sparse
>>> sparse.random([1000000,1000000,1000000,1000000], nnz=4)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/willow/Library/Caches/pypoetry/virtualenvs/willow-ENU-A1do-py3.11/lib/python3.11/site-packages/sparse/_utils.py", line 340, in random
    ).reshape(shape)
      ^^^^^^^^^^^^^^
  File "/Users/willow/Library/Caches/pypoetry/virtualenvs/willow-ENU-A1do-py3.11/lib/python3.11/site-packages/sparse/_coo/core.py", line 1029, in reshape
    raise ValueError(f"cannot reshape array of size {self.size} into shape {shape}")
ValueError: cannot reshape array of size 2003764205206896640 into shape (1000000, 1000000, 1000000, 1000000)
```

willow-ahrens:
Just fyi, that number is the overflowed version of 1000000^4

```
julia> t = 1000000
1000000

julia> t * t * t * t
2003764205206896640
```

willow-ahrens:
Does this look good to everyone? There's a lot of statistics here so it would be nice if someone else could review.

mtsokol (approved):

hameerabbasi:
> Just fyi, that number is the overflowed version of 1000000^4
> 
> 
> 
> ```
> 
> julia> t = 1000000
> 
> 1000000
> 
> 
> 
> julia> t * t * t * t
> 
> 2003764205206896640
> 
> ```

Yes, in `sparse`, we currently don't support arrays of size `>= 2 ** 64`.

PR #497, opened 2024-04-15
Add compat entry for `Distributions`

mtsokol:
Hi @willow-ahrens,

Registration of the latest Finch version failed due to missing `[compat]` entry:

https://github.com/JuliaRegistries/General/pull/104928

I think this change should fix it.


willow-ahrens:
There's a bot that makes these prs for us, I just hadn't merged it yet! Its on its way now, so we can retry registration in a bit

willow-ahrens:
I realize this pr is the same as the bot, so I just merged this one

mtsokol:
Ok, I think the new version got pushed to the julia registry.
But this time the tag and release didn't get created automatically.

PR #512, opened 2024-04-29
Fix empty COO segfault

### LINKED ISSUE #511: Segfault when converting COO to Dense
Hi @willow-ahrens,

I'm trying to convert empty COO tensor to a dense one but I get a segfault. Should I write it differently or is it a bug?

```julia
using Finch

tns_coo = Tensor(SparseCOO{2}(Element(0), (3,3)))  # empty coo

Tensor(Dense(Dense(Element(0))), b)  # segfault instead of dense 3x3
```

mtsokol:
Fixes #511

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Finch"
 uuid = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
 authors = ["Willow Ahrens"]
-version = "0.6.23"
+version = "0.6.24"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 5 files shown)


willow-ahrens:
I don't think this function was supposed to create an initialized tensor, and corresponding versions in other levels also will not. This also won't fix the case where we have `Dense(SparseCOO(...))`, since you would need an array of `n` ones. I think it might be misleading to try to fix just this one case, so I'd rather close this if you're okay with it.

Zooming out, I'm okay with changing the tensor constructors to always initialize tensors, but it is often useful to be able to construct a tensor from it's fields without initializing it.

mtsokol:
Thank you for an explanation - now I have the right semantics.

PR #515, opened 2024-04-30
add atomic element level

### LINKED ISSUE #531: Documentation needed for separated memory level and Atomic Level
DenseRLELevel, SparseHash, Separate, SparseBand, SparsePoint, and SparseVBL formats currently lack support for parallel SpMV.

willow-ahrens:
Co-authored-by: teocollin1995 <teoc@mit.edu>, sortof fixes #531


willow-ahrens:
We should add a lower_assign function here.

willow-ahrens:
thus, instantiate would be an "early expand" function, that expands levels immediate after unfurling. "unfurl" and "lower_access" would be "late expand" functions, that don't add dimensions and wait until the next call to unfurl to expand themselves.

src/tensors/levels/mutex_levels.jl
```diff
@@ -0,0 +1,272 @@
...
+    return lvl
+end
+
+function virtual_moveto_level(ctx::AbstractCompiler, lvl::VirtualMutexLevel, arch)
+    #Add for seperation level too.
+    atomics = freshen(ctx, :locksArray)
+
+    push_preamble!(ctx, quote
+        $atomics = $(lvl.locks)
+        $(lvl.locks) = $moveto($(lvl.locks), $(ctx(arch)))
```
wraith1995, line 188: I know I probably wrote something like this in the initial implementation but something about this does strike me as kind of cursed given the intended use of moveto.

willow-ahrens: Ah, yes. I recently fixed #608, and I imagine at some point we will need to get a little more precise about how we use moveto (perhaps explicitly making copies of state variables and other things)

src/tensors/levels/mutex_levels.jl
```diff
@@ -0,0 +1,272 @@
...
+        $lockVal = Finch.aquire_lock!($dev, $atomicData)
+    end)
+    res = lower_assign(ctx, VirtualSubFiber(lvl.lvl, pos), mode, op, rhs)
+    push_epilogue!(ctx, quote
+        Finch.release_lock!($dev, $atomicData)
+    end)
+    return res
+end
+
+function lower_assign(ctx, fbr::VirtualHollowSubFiber{VirtualMutexLevel}, mode::Updater, op, rhs)
```
wraith1995, line 257: one thing that I do like about this function is now we could add a way to do an atomicOp via a sort of "HollowAtomic Level", which translates certain ops to atomic operations depending on what the arch supports.

willow-ahrens: Yes, that was the intention, currently that stuff lives in architecture.jl

src/tensors/levels/sparse_list_levels.jl
```diff
@@ -141,8 +141,8 @@ end
 
 is_level_injective(ctx, lvl::VirtualSparseListLevel) = [is_level_injective(ctx, lvl.lvl)..., false]
 function is_level_atomic(ctx, lvl::VirtualSparseListLevel)
-    (below, atomic) = is_level_atomic(ctx, lvl.lvl)
-    return ([below; [atomic]], atomic)
+    (below, Mutex) = is_level_atomic(ctx, lvl.lvl)
```
wraith1995, line 144: I forget how this works, but shouldn't a false be being added somewhere here?

wraith1995: and in many of these.

wraith1995 (commented):
Minor comments + one concern.

willow-ahrens:
I'm worried that mutex levels can't handle multiple writers to the same level from the same thread. Not sure how we would modify them to support that. I guess we could check and set the lock to the thread id, and wait for the tid to be the same.

src/tensors/levels/sparse_list_levels.jl
willow-ahrens: I honestly don't know why this changed.

willow-ahrens: This is how it was before we started. Perhaps I need to make an issue for this. I'll look into it. One related question I had for you: This PR delays the lowering of atomics until just before their child is unfurled or accessed. That doesn't change this analysis right?

wraith1995: That is going to require some thought for me to answer. Hmm.

wraith1995:
> I'm worried that mutex levels can't handle multiple writers to the same level from the same thread. Not sure how we would modify them to support that. I guess we could check and set the lock to the thread id, and wait for the tid to be the same.

I think we don't want the various atomic levels to have knowledge of the thread id. That strikes me as something we should avoid.

I think the concern here is a deadlock, right? I think the main thing there is just this is inefficient because we need to release the lock after each write. The key concern is where the compiler places the acquire/releases. IIRC we wanted atomic to bind as tightly as possible - if that is the case, I think this should be fine.

willow-ahrens:
if we write
```
for i = _
   A[i] += 1
   A[i] += 2
end
```
and A is Tensor(Atomic(Dense(Element(0.0)))), it seems like this is a case where we would be okay if we knew that it was the same thread doing both modifications.

willow-ahrens:
I guess what I'm saying is that we want that case to be okay if only one thread is running this computation, but it would currently deadlock.

wraith1995:
> if we write
> 
> ```
> for i = _
>    A[i] += 1
>    A[i] += 2
> end
> ```
> 
> and A is Tensor(Mutex(Dense(Element(0.0)))), it seems like this is a case where we would be okay if we knew that it was the same thread doing both modifications.

I have a test meant to stress atomics - let's adapt that to this? Then I think the next thing to do is ask how we can either ensure that releases come before acquires or else ensure both are grouped under one lock. We need one of those, correct?

willow-ahrens:
The compiler unwraps both atomic accesses before lowering the loop. This means it lowers to
```
aquire
aquire
for i
  ...
end
release
release
```

PR #542, opened 2024-05-10
Example `finch-tensor` tests

mtsokol:
Not for merging.


willow-ahrens:
We'll also need guidelines in contributing.md on how to run the python tests.

PR #543, opened 2024-05-10
hotfix einsum

willow-ahrens:
fixes einsum for scalars


willow-ahrens:
@albertlu03 after this merges i'll register finch 0.6.26

PR #547, opened 2024-05-10
Kbd-wrapperize-fix

kylebd99:
Fixes issue @546 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Finch"
 uuid = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
 authors = ["Willow Ahrens"]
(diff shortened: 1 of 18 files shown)


willow-ahrens (commented):
While we're at it, I'd like to make some additional fixes to other calls in wrapperize.jl

willow-ahrens:
Okay, I've made my fixes. I thought I would need to call getval somewhere else in that file but I didn't. That's because the rules for other wrappers like offset don't need to inspect the literal values, they just wrap them in `call` nodes.

willow-ahrens:
There's still a fair amount of uncovered code in wrapperize.jl, but this PR is a strict improvement so we'll merge it. i think we need to start pushing for more code coverage though.

PR #551, opened 2024-05-12
Fix `fspzeros` docstring signature to match implementation

ericphanson:


willow-ahrens:
Thanks! This looks good to me.

Also, a question: If Finch were to instead overload `SparseArrays.sprand` anytime a user provided a tuple of dimensions, would that feel more discoverable?

ericphanson:
I personally like the separate functions; I think I would find that overload confusing, since the tuple-vs-not distinction is kind of small and doesn't really scream "Finch" to me

Issue #537, opened 2024-05-09
binsparse is out of date


willow-ahrens:
https://github.com/GraphBLAS/binsparse-specification/pull/52/files

Issue #523, opened 2024-05-05
matrixmarket reader may be swapping row and column


willow-ahrens:
https://github.com/GraphBLAS/binsparse-specification/pull/39

PR #555, opened 2024-05-13
fixing binsparse

willow-ahrens:
fixes https://github.com/willow-ahrens/Finch.jl/issues/523 fixes https://github.com/willow-ahrens/Finch.jl/issues/537


willow-ahrens:
Also need to rename swizzle->transpose

PR #561, opened 2024-05-16
Wma/parallel improvements

willow-ahrens:


willow-ahrens:
The problem here is that we're trying to use FastClosures, let's move to just use polyester for now, and make an issue for fastclosures (which is the better overall option going forward)

PR #569, opened 2024-05-22
Exponential Scansearch

kylebd99:
Small update to the scansearch function which replaces the linear + binary search with an exponential search algorithm.


willow-ahrens:
oh cool! just a small comment here: There may be a few situations (e.g. a follow protocol on a sparse list) where we assume the destination is random, not nearby, so we may want to leave a binary search in the codebase as well.

kylebd99:
Added a binary search back in as binary_scansearch!

PR #581, opened 2024-05-29
SparseByteMap follow fix

### LINKED ISSUE #580: SparseByteMap Follow Protocol Incorrect Results
From what I can tell, the sparse byte map level zeros out when you use the follow protocol producing incorrect results.

```
using Finch

n = 100
data =  fsprand(Int, n, n, .01)
A_sbm = Tensor(SparseByteMap(SparseByteMap(Element(0))), data)
A_d = Tensor(Dense(Dense(Element(0))), data)

C1 = Scalar(0)
C2 = Scalar(0)

@finch begin
    C1 .= 0
    for i=_
         for j=_
             for k=_
                 C1[] += A_sbm[follow(j),i] * A_sbm[k,j]
             end
         end
     end
     C2 .= 0
     for i=_
          for j=_
              for k=_
                C2[] += A_d[j,i] * A_d[k,j]
              end
          end
      end
end

println(C1)
println(C2)
```

kylebd99:
Fix SparseByteMap follow 

fixes #580

### DIFF
--- a/src/tensors/levels/sparse_bytemap_levels.jl
+++ b/src/tensors/levels/sparse_bytemap_levels.jl
@@ -5,7 +5,7 @@ Like the [`SparseListLevel`](@ref), but a dense bitmap is used to encode
 which slices are stored. This allows the ByteMap level to support random access.
 
 `Ti` is the type of the last tensor index, and `Tp` is the type used for
-positions in the level. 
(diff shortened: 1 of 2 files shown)


willow-ahrens (approved):
Thanks for the fix!

PR #587, opened 2024-06-07
perf improvement

willow-ahrens:
should improve compile time a bit?


willow-ahrens:
by my measurements, this looks like a 2.5x improvement in @finch_code and @finch_kernel runtime, and a 1.1x improvement in execute_code

PR #600, opened 2024-06-13
Improve the Sparse level's sort procedure

kylebd99:
This uses a two-step sort where we first use tbl.ptr to bucket the idxs & values. Then, we sort each bucket individually. Some initial testing showed a 2x matrix transpose improvement:
```
using Finch
using BenchmarkTools


n = 100000
nnz = 1000000
A = Tensor(Dense(SparseList(Element(0.0))), fsprand(n, n, nnz))
C = Tensor(Dense(Sparse(Element(0.0))))

eval(@finch_kernel mode=:fast function tp(A, C)
    C .= 0
    for i=_
        for j=_
            C[i,j] = A[j,i]
        end
    end
end)

hash_times = []

@benchmark tp($A, $C)
```

### DIFF
--- a/src/tensors/levels/sparse_levels.jl
+++ b/src/tensors/levels/sparse_levels.jl
@@ -32,6 +32,24 @@ function assemble_table!(tbl::DictTable, pos_start, pos_stop)
     fill_range!(tbl.ptr, 0, pos_start + 1, pos_stop + 1)
 end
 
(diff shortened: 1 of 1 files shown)


willow-ahrens:
this looks good! We can merge this now, and I'll see if I can tweak it later.

PR #606, opened 2024-06-18
Release 0.6.32

mtsokol:
Hi @willow-ahrens,

Let's create a new release once we're ready. In `finch-tensor` and locally the dev mode works fine, but we accumulated some pending changes in `sparse` https://github.com/pydata/sparse/pull/702 (mostly solved Array API test xfails) that require a new `finch-tensor` release with new `Finch.jl`.


willow-ahrens:
Sure, go ahead with a new release! I'm hoping to add some additional documentation before 0.7, so let's make this a patch release 0.6.32

PR #611, opened 2024-07-12
Created an example using Finch 

ecoArcGaming:
Just a short graph algorithm that finds the number of connected components given an undirected adjacency matrix.


willow-ahrens:
This is cool! Unfortunately, I don't think Finch supports this kind of syntax, but it would be cool if it did. In the meantime, I think we could probably implement this algorithm for connected components:

https://people.eecs.berkeley.edu/~aydin/LACC.pdf

willow-ahrens:
I'm closing this for now, but feel free to push updates and reopen if you like.

PR #612, opened 2024-08-01
Create topological sorting example

Paramuths:
Implementing topological sorting example using finch


willow-ahrens (changes requested):
This PR looks great, but I'm hesitant to add it as an example unless we make the tensors sparse and test that it's correct and performant.

PR #620, opened 2024-10-21
Update ⚙ in tensor_formats.md for consistency

AbdAlazezAhmed:
I noticed that there's that one gear icon that's different than the rest, so here's a one character diff pr :D
Also, the link in the about section seems to be outdated.


willow-ahrens (approved):
Thanks!

PR #634, opened 2024-11-08
Introduce a de-parser to return Finch expressions to their readable macro input form

### LINKED ISSUE #624: FinchProgram => @finch invocation
We need the reverse parser sometimes, the relevant code is in syntax.jl

willow-ahrens:
fixes #624


kylebd99:
This looks great to me! I think the tests might need to be updated to make check_outputs work. But, I've used it in Galley, and it passed tests, so I think it's solid.

kylebd99 (approved):

PR #636, opened 2024-11-12
Kbd small finch logic changes

kylebd99:
Builds on the de-parser #634 and makes some small changes to the deferred node. In particular, it includes the tensor instance in deferred so that optimizers can access the data to gather statistics.

### DIFF
--- a/src/FinchLogic/nodes.jl
+++ b/src/FinchLogic/nodes.jl
@@ -205,6 +205,8 @@ function LogicNode(kind::LogicNodeKind, args::Vector)
         return LogicNode(kind, args[1], Any, LogicNode[])
     elseif kind === deferred && length(args) == 2
         return LogicNode(kind, args[1], args[2], LogicNode[])
+    elseif kind === deferred && length(args) == 3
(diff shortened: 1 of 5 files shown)


willow-ahrens (changes requested):
This looks good to me! There's no reason not to include the tensors in the deferred nodes. I've made a few changes to get this to pass tests. I'm curious about testing. Should we add a CI script to run galley tests, similar to how we also test finch changes against python downstream? Similarly, we might want to figure out a simple test for the finch_unparse function in the Finch codebase itself. what do you think that would look like? I suppose we could do
```
prgm = @finch_program for i = _
   A[i] += 1
end
eval(quote
   @finch $(finch_unparse_program(prgm))
end)
```
to test that the roundtrip works.

kylebd99:
I just added a test for finch_unparse_program. Galley relies on this function now, so I'm going to merge this PR before adding a PR with CI for Galley. Just to avoid a circular reference situation.

willow-ahrens (approved):

PR #638, opened 2024-11-15
Version Bump to "0.6.33"

kylebd99:
Lot's of things have changed since June. So, it might be nice to do a new release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Finch"
 uuid = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
 authors = ["Willow Ahrens"]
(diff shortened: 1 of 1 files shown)


willow-ahrens:
good call. We have several bug fixes, etc. At some point in the near future, after I figure out the documentation issue, it might be time for a 1.0

willow-ahrens:
At the very least, it would be good to get serious about semver

willow-ahrens:
@kylebd99 any idea why galley CI would fail here but not on the last pr?

kylebd99:
That's my bad. I changed the test dependencies for Galley and it broke the CI. Just fixed it over in the Galley repo, so it should pass now.

PR #641, opened 2024-11-18
Add Galley into Finch Codebase

kylebd99:
In order to ease downstream integrations, this PR adds the codebase from Galley.jl into the Finch.jl codebase as a sub-module. This means that all users which call `using Finch` will have access to Galley as an optimizer without additional deps. This should simplify the user experience and allow for the possibility of Galley as a default in the future.

Necessary Changes:
- [x] Remove galley backtests from CI
- [x] Add galley sub-module
- [x] Merge galley tests into one file and add to runtests.jl
- [x] Add galley deps to project.toml
- [x] Add documentation for Galley in Array API


src/Galley/TensorStats/tensor-stats.jl
willow-ahrens, line 1: Is this file not used?

src/Galley/TensorStats/propagate-stats.jl
willow-ahrens, line 1: Is this file also not used?

willow-ahrens:
The documentation in the array API section is likely to get moved around in https://github.com/finch-tensor/Finch.jl/pull/640

willow-ahrens:
Perhaps we can merge both of these, then do an update which improves the docs.

src/Galley/TensorStats/propagate-stats.jl
kylebd99: These are both definitely used. I just left some old commented out includes. They're gone now.

willow-ahrens (approved):
This looks good to me. There's a lot of code here, but as this passes tests, we can merge here and sort out the fine tuning later.

willow-ahrens:
These windows failures look like they are from expecting UInt == UInt64 somewhere, or similar, but I can't find where

test/test_galley.jl
```diff
@@ -0,0 +1,591 @@
...
+        c_data = Tensor(SparseList(SparseList(Element(0), 100), 100))
+        copyto!(c_data, c_matrix)
+        c = Input(c_data, :k, :l)
+        d = Aggregate(+, 0, :j, MapJoin(*, a, b))
+        e = Query(:out, Materialize(t_sparse_list, t_sparse_list, :i, :l, Aggregate(+, 0, :k, MapJoin(*, d, c))))
+        result = galley(e, verbose=verbose)
+        d_matrix = a_matrix * b_matrix
+        correct_matrix = d_matrix * c_matrix
+        @test result.value[1] == correct_matrix
+    end
```
willow-ahrens, line 228: This is the last failing test.

PR #649, opened 2024-11-21
add instance node

### LINKED ISSUE #647: Tag FinchLogicExecutor

kylebd99:
Attempt to fix #647 . This PR adds an "instance" node which is supposed to identify a program as a unique compilation instance. Effectively, this is a promise that later invocations will be with similar data, so should benefit from similar compiler optimizations.


willow-ahrens:
Does it make more sense to add this as an argument to the scheduler?

kylebd99:
Like an argument to FinchLogicExecutor()?

willow-ahrens:
This looks like the right move. Can we rename "instance" to something like "tag"? "Instance" is a little overloaded in this codebase

willow-ahrens:
in line https://github.com/finch-tensor/Finch.jl/blob/730cf6bef9240714b092b7d38ff7fbd3a1c9e76a/src/interface/lazy.jl#L521-L523, we pass toplevel kwargs to optimizers, it's how we pass verbose all the way through the scheduler interface.

kylebd99:
That makes sense, I just made the tag parameter a part of the LogicExecutor struct, and it's now passed through set_options. This seems more in line with how the other params are handled.

willow-ahrens:
I merged main, but I can't help but wonder if instead we should be caching scheduler state along with the tag as well. I'm not really sure how to do it, but the executor would be the place to do so.

willow-ahrens:
For example, if we want galley to remember the last few inputs we sent to it, we would want to keep some sort of galley state around, associated with the tag.

PR #657, opened 2024-11-28
Replace pyjulia with juliacall in docs

gdalle:
The README of pyjulia itself points to juliacall so I think it's warranted


willow-ahrens (approved):
I made a few small additional edits, thanks for updating this!

PR #670, opened 2024-12-11
Tensor Init Bug Fix

kylebd99:
Galley's tensor initialization is improper, leading to memory leaks between invocations. 

This patch fixes that by updating the init to reference an expression rather than a tensor.


willow-ahrens:
This looks good to me, can you verify that it fixes mateusz's issue?

kylebd99:
Can confirm. Ran it with main and it crashes, ran it with this and it doesn't.

PR #687, opened 2025-01-05
Update `propagate_map_queries`

mtsokol:
Hi @willow-ahrens,

I started rewriting default scheduler (for now to Python) and I noticed that `propagate_map_queries` leaves empty plans after removing queries. Here's an example that reproduces it:
```julia
julia> plan = Finch.plan(Finch.query(Finch.alias(:A), Finch.aggregate(Finch.immediate(:B), Finch.immediate(:C), Finch.immediate(:D))), Finch.query(Finch.alias(:E), Finch.alias(:A)), Finch.produces(Finch.alias(:E)))
Finch Logic: plan
  A = aggregate(B, C, D)
  E = A
  return (E)
end

julia> res = Finch.propagate_map_queries(plan)
Finch Logic: plan
  plan
  end
  E = mapjoin(B, C, D)
  return (E)
end
```

It occurs because the third rule at the end of `propagate_map_queries` is never applied due to the Prewalk traversal - top level `Plan` is visited first so the empty/stub `Plan()` can't be removed.

My proposition is to make the third rule a separate `Rewrite`. WDYT?

### DIFF
--- a/src/scheduler/optimize.jl
+++ b/src/scheduler/optimize.jl
@@ -410,11 +410,11 @@ function propagate_map_queries(root)
             end
         end
     end
-    Rewrite(Prewalk(Chain([
+    root = Rewrite(Prewalk(Chain([
         (a -> if haskey(props, a) props[a] end),
         (@rule query(~a, ~b) => if haskey(props, a) plan() end),
-        (@rule plan(~a1..., plan(), ~a2...) => plan(a1..., a2...)),
     ])))(root)
+    Rewrite(Postwalk(@rule plan(~a1..., plan(), ~a2...) => plan(a1..., a2...)))(root)
 end
 
 function propagate_map_queries_backward(root)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -119,6 +119,7 @@ include("utils.jl")
         end
     end
     if should_run("parallel") include("test_parallel.jl") end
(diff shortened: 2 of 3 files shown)


mtsokol:
I also noticed that in [the first line of `propagate_map_queries`](https://github.com/finch-tensor/Finch.jl/blob/061cda793e3adb006fb6518f5ed14c7e113b81e2/src/scheduler/optimize.jl#L403) `aggregate` is missing `idxs...` argument. Is it intended? (I'm not fully familiar with the rewrite mechanism yet)

willow-ahrens (approved):
Ah, good catch! It's not required to remove empty plans, it just makes the program prettier (they are no-ops, so they don't affect the meaning of the program).

willow-ahrens:
The intention of `root = Rewrite(Postwalk(@rule aggregate(~op, ~init, ~arg) => mapjoin(op, init, arg)))(root)` is to convert "aggregates" which reduce over no variables into their equivalent mapjoins. The line would be incorrect if there were indices in the aggregate we were supposed to reduce over, as the aggregate would no longer be equivalent to the mapjoin. If you're working with the default scheduler, you may want to take a look at https://github.com/willow-ahrens/RewriteTools.jl, especially the docstrings (the package doesn't have fancy docs like Finch but the docstrings are good).

PR #694, opened 2025-01-17
Adding stats functions

### LINKED ISSUE #574: `mean`, `std`, and `var` functions
Hi @willow-ahrens,

This is a tracking issue for [mean](https://data-apis.org/array-api/latest/API_specification/generated/array_api.mean.html), [std](https://data-apis.org/array-api/latest/API_specification/generated/array_api.std.html), and [var](https://data-apis.org/array-api/latest/API_specification/generated/array_api.var.html) functions for `SwizzleArray`, `LazyArray`, and `Tensor` objects (similar to `max`, `min` that we already have).

no-dpr:
fixes #574 

Implementing size property to lazy tensors to use with stats functions.


src/interface/lazy.jl
```diff
@@ -7,12 +7,13 @@ const AbstractArrayOrBroadcasted = Union{AbstractArray,Broadcasted}
 mutable struct LazyTensor{T, N}
     data
     extrude::NTuple{N, Bool}
+    size::NTuple{N, UInt64}
```
willow-ahrens, line 10: This field should be renamed `shape` for consistency with other naming choices in the codebase

willow-ahrens (changes requested):
please rename the field

src/Finch.jl
```diff
@@ -54,6 +54,7 @@ export lazy, compute, fused, tensordot, @einsum
 export choose, minby, maxby, overwrite, initwrite, filterop, d
 
 export fill_value, AsArray, expanddims, tensor_tree
+export mean
```
willow-ahrens, line 57: instead of exporting our own mean, we should use the `mean` from the `Statistics` standard library. We also don't want to depend on `Statistics` explicitly, instead we should use an "optional dependency", where `mean`, `std`, and `var` definitions are loaded only when the user loads both `Finch` and `Statistics`. `Finch` currently takes this approach with `SparseArrays`. You can see the approach in [`ext/SparseArraysExt.jl`](https://github.com/finch-tensor/Finch.jl/blob/main/ext/SparseArraysExt.jl). We would also need to handle `Statistics` the same way we handle `SparseArrays` in the [`Project.toml`](https://github.com/finch-tensor/Finch.jl/blob/main/Project.toml). More on this concept here: https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

src/interface/eager.jl
```diff
@@ -112,14 +112,25 @@ Base.maximum(arr::AbstractTensorOrBroadcast; kwargs...) = reduce(max, arr; init
...
 Base.extrema(arr::AbstractTensorOrBroadcast; kwargs...) = mapreduce(plex, min1max2, arr; init = (typemax(broadcast_to_eltype(arr)), typemin(broadcast_to_eltype(arr))), kwargs...)
 
+
 function LinearAlgebra.norm(arr::AbstractTensorOrBroadcast, p::Real = 2)
     compute(norm(lazy(arr), p))[]
 end
 
 """
-    expanddims(arr::AbstractTensor, dims)
+expanddims(arr::AbstractTensor, dims)
```
willow-ahrens, line 121: This line should stay indented

src/interface/lazy.jl
```diff
@@ -7,12 +7,13 @@ const AbstractArrayOrBroadcasted = Union{AbstractArray,Broadcasted}
 mutable struct LazyTensor{T, N}
     data
     extrude::NTuple{N, Bool}
+    shape::NTuple{N, Int}
```
willow-ahrens, line 10: For better or worse, Finch array dimensions may use different types independently. So you could have a matrix with row dimension `UInt8` and column dimension `BigInt`. The way we handle this is to parameterize the type of LazyTensor based on the type of the dimensions: 
```
mutable struct LazyTensor{T, N, Shape <: Tuple}
    data
    extrude::NTuple{N, Bool}
    shape::Shape
    fill_value::T
end
```
Note that the dimension tuple is not constrained to be length `N` with the type annotations. That's okay, we can check that it's correct when we construct the LazyTensor.

src/interface/lazy.jl
```diff
@@ -7,12 +7,13 @@ const AbstractArrayOrBroadcasted = Union{AbstractArray,Broadcasted}
...
     fill_value::T
 end
-LazyTensor{T}(data, extrude::NTuple{N, Bool}, fill_value) where {T, N} = LazyTensor{T, N}(data, extrude, fill_value)
+LazyTensor{T}(data, extrude::NTuple{N, Bool}, shape::NTuple{N, Int}, fill_value) where {T, N} = LazyTensor{T, N}(data, extrude, shape, fill_value)
 
 function Base.show(io::IO, tns::LazyTensor)
-    join(io, [x ? "1" : "?" for x in tns.extrude], "×")
+    join(io, [string(x) for x in tns.shape], "×")
```
willow-ahrens, line 16: I think you may also be able to get away with `join(io, tns.shape, "×")`, but I'm not sure, try it?

src/interface/lazy.jl
```diff
@@ -103,12 +113,12 @@ function Base.map(f, src::LazyTensor, args...)
...
     new_default = f(map(fill_value, largs)...)
     data = mapjoin(immediate(f), ldatas...)
-    return LazyTensor{T}(identify(data), src.extrude, new_default)
+    return LazyTensor{T}(identify(data), src.extrude, src.shape, new_default)
 end
 
 function Base.map!(dst, f, src::LazyTensor, args...)
     res = map(f, src, args...)
-    return LazyTensor(identify(reformat(dst, res.data)), res.extrude, res.fill_value)
+    return LazyTensor(identify(reformat(dst, res.data)), res.extrude, res.size, res.fill_value)
```
willow-ahrens, line 121: Almost certainly, this should be `size(res)` or `res.shape`, depending on whether or not we know that `res` is a `LazyTensor`.

willow-ahrens (changes requested):
Great work! This is looking good! I've added some comments about packaging. This PR needs some tests as well. You could add tests to interface_tests.jl. Be sure to leave a link to the github issue to document what precisely the tests are testing.

willow-ahrens:
Let's also check that the shape of the lazy tensors is always equal to the shape of the returned tensors when in debug mode using an `@assert` statement in `@debug` mode [here](https://github.com/finch-tensor/Finch.jl/blob/b5a1a87d49365ab6c694d472f72e22586cbd9ce0/src/interface/lazy.jl#L637)

willow-ahrens (approved):

src/Galley/FinchCompat/translate.jl
```diff
@@ -111,7 +109,8 @@ function remove_reorders(prgm::LogicNode)
                         if length(idxs2) < length(getfields(arg))
                             reorder(
                                 aggregate(
-                                    nothing, nothing, arg, setdiff(getfields(arg), idxs2)...
+                                    overwrite, nothing, arg,
```
willow-ahrens, line 112: @kylebd99, I found this was necessary to get things to work, and I'm not sure why. The issue was that merge_stats_def would try to evaluate the aggregate function because the original rule to remove `isnothing(init)` aggregates never matched. I'm just confused about how we never hit this before.

kylebd99: Hmm, yeah I don't have a lot of intuition on this one. It seems like the original version should have checked isnothing on `op` not `init`. But, I'm quite surprised that it was never firing.

PR #699, opened 2025-02-10
Add DFS and Degree Centrality Examples

01110011011101010110010001101111:
This adds two new examples: DFS and Degree Centrality.

Degree Centrality is relatively straightforward (the sum of the indegree and outdegree normalized by the number of vertices - 1).

DFS is also relatively straightforward, can be room for performance improvement with the stack concatenation. 

- [x] Added two new examples
- [x] Added corresponding tests
- [x] Verified rest of tests (namely style) still pass


willow-ahrens (approved):
Let's merge these examples!

PR #705, opened 2025-02-23
Support for argmin / argmax

### LINKED ISSUE #701: Implement argmin and argmax
The idea is to first use minby to compute the argmin, then return just the coordinates of the minimum. In python, we only need to handle a single axis (https://numpy.org/doc/2.1/reference/generated/numpy.argmin.html). In Julia, there may be multiple axes, in which case the index used is an e.g. `CartesianIndex(i, j)` (https://docs.julialang.org/en/v1/base/collections/#Base.argmin). The `minby` function is described here: https://finch-tensor.github.io/Finch.jl/stable/docs/user-defined_functions/#Finch.minby

An example of a one-dimensional argmin is:

```
A = Scalar((Inf, 0))
@finch begin
    A .= (Inf, 0)
    for i = _
      A[] <<minby>>= tuple(B[i], i)
   end
end
result = A[][2]
```

An example of a similar function exists here: hhttps://github.com/finch-tensor/Finch.jl/blob/2e89d2228a777a8238eccd145cfa6a7b405eb76c/src/interface/compare.jl#L1-L28

You may find some of these docs to be helpful:
https://finch-tensor.github.io/Finch.jl/stable/docs/internals/virtualization/

01110011011101010110010001101111:
Closes #701 fixes https://github.com/finch-tensor/Finch.jl/issues/724

Uses similar setup to `isqqual` and the example code for the 1D `argmin`/`argmax`. Similar to the `argmin`/`argmax` in Julia, uses CartesianIndex for N-d tensors. 

Tested for 1d and 2d matrices.


src/interface/compare.jl
```diff
@@ -55,3 +55,51 @@ end
...
+@staged function helper_argmin(A)
+   idxs = [Symbol(:i_, n) for n in 1:ndims(A)]
+   exts = Expr(:block, (:($idx = _) for idx in reverse(idxs))...)
+   return quote
+       x = Scalar(Inf => CartesianIndex(zeros(Int, ndims(A))...))
+       @finch $(Expr(
+           :for,
+           exts,
+           quote
+               x[] <<minby>>= A[$(idxs...)] => CartesianIndex($(idxs...))
```
willow-ahrens, line 68: This looks great! When the vector is one-dimensional (or when we call from python with a dim equal to an integer), I think we also need a version that returns integer outputs, rather than cartesianIndex. Also, when dims is omitted, we return a cartesianindex instead of an array of cartesianindex

willow-ahrens (changes requested):
We don't need to handle the python case (though it would be nice to handle it as a separate function `argmin_python`.) We do need to handle the case where dims is left out. In that case we unwrap the output, so instead of returning `[CartesianIndex(3, 4)]` we return `CartesianIndex(3, 4)`. We also need to handle the vector case.

src/interface/compare.jl
willow-ahrens: Oops! I think I misread your code as supporting the version with `dims`. Sorry for the confusion! In any case, we should support both versions. In Julia, you can access help mode with the ? at the repl prompt. The relevant documentation for argmin reads:

```
help?> argmin
search: argmin argmax min begin asin hardlink atomic

  argmin(itr)

  Return the index or key of the minimal element in a collection. If there are multiple minimal elements, then the first one will be returned.

  The collection must not be empty.

  Indices are of the same type as those returned by keys(itr) and pairs(itr).

  NaN is treated as less than all other values except missing.

  See also: argmax, findmin.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> argmin([8, 0.1, -9, pi])
  3
  
  julia> argmin([7, 1, 1, 6])
  2
  
  julia> argmin([7, 1, 1, NaN])
  4

  ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  argmin(A; dims) -> indices

  For an array input, return the indices of the minimum elements over the given dimensions. NaN is treated as less than all other values except missing.

  Examples
  ≡≡≡≡≡≡≡≡

  julia> A = [1.0 2; 3 4]
  2×2 Matrix{Float64}:
   1.0  2.0
   3.0  4.0
  
  julia> argmin(A, dims=1)
  1×2 Matrix{CartesianIndex{2}}:
   CartesianIndex(1, 1)  CartesianIndex(1, 2)
  
  julia> argmin(A, dims=2)
  2×1 Matrix{CartesianIndex{2}}:
   CartesianIndex(1, 1)
   CartesianIndex(2, 1)
```
When dims is specified (`argmin(A, dims=2)`), we return an array of indices. When `dims` is absent (`argmin(itr)`) we return a single index. When the array is two or more dimensional, we use CartesianIndex. Otherwise we use a regular integer.

For the python version (`argmin_python`):
https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.argmin.html
When dim is specified it will only be a single dimension and we use integer indices with that dimension.
When dim is absent, we return an integer index into the "flattened" array, meaning we convert the CartesianIndex into a linear index by multiplying the indices by the strides (`i * n + j`).

01110011011101010110010001101111:
Finally updated. There's a lot of redundancy so happy to take recommendations to reduce that. Example of all workflows (can also tests as needed): 

```
julia> A = fsprand(5, 3, 0.8)
5×3 Tensor{SparseCOOLevel{2, Tuple{Int64, Int64}, Vector{Int64}, Tuple{Vector{Int64}, Vector{Int64}}, ElementLevel{0.0, Float64, Int64, Vector{Float64}}}}:
 0.49954   0.235891  0.0
 0.0       0.285554  0.411685
 0.704496  0.29967   0.791234
 0.988549  0.0       0.557997
 0.959831  0.0       0.738574

julia> argmin(A, dims=1)
3 Tensor{DenseLevel{Int64, ElementLevel{CartesianIndex(0, 0), CartesianIndex{2}, Int64, Vector{CartesianIndex{2}}}}}:
 CartesianIndex(2, 1)
 CartesianIndex(4, 2)
 CartesianIndex(1, 3)

julia> argmin(A, dims=2)
5 Tensor{DenseLevel{Int64, ElementLevel{CartesianIndex(0, 0), CartesianIndex{2}, Int64, Vector{CartesianIndex{2}}}}}:
 CartesianIndex(1, 3)
 CartesianIndex(2, 1)
 CartesianIndex(3, 2)
 CartesianIndex(4, 2)
 CartesianIndex(5, 2)

julia> argmax(A, dims=1)
3 Tensor{DenseLevel{Int64, ElementLevel{CartesianIndex(0, 0), CartesianIndex{2}, Int64, Vector{CartesianIndex{2}}}}}:
 CartesianIndex(4, 1)
 CartesianIndex(3, 2)
 CartesianIndex(3, 3)

julia> argmax(A, dims=2)
5 Tensor{DenseLevel{Int64, ElementLevel{CartesianIndex(0, 0), CartesianIndex{2}, Int64, Vector{CartesianIndex{2}}}}}:
 CartesianIndex(1, 1)
 CartesianIndex(2, 3)
 CartesianIndex(3, 3)
 CartesianIndex(4, 1)
 CartesianIndex(5, 1)

julia> argmax(A)
CartesianIndex(4, 1)

julia> argmin(A)
CartesianIndex(2, 1)

julia> A = fsprand(5, 0.8)
5 Tensor{SparseCOOLevel{1, Tuple{Int64}, Vector{Int64}, Tuple{Vector{Int64}}, ElementLevel{0.0, Float64, Int64, Vector{Float64}}}}:
 0.6417627318700393
 0.9963457229662037
 0.0
 0.41044352897956116
 0.2289561999682863

julia> argmax(A)
2

julia> argmax(A, dims=1)
 Tensor{ElementLevel{0, Int64, Int64, Vector{Int64}}}:
2

julia> argmin(A)
3

julia> argmin(A, dims=1)
 Tensor{ElementLevel{0, Int64, Int64, Vector{Int64}}}:
3
```

src/interface/compare.jl
```diff
@@ -55,3 +55,69 @@ end
 function Base.isequal(A::AbstractArray, B::AbstractTensor)
     return helper_isequal(A, B)
 end
+
+function helper_argmin(A, dims)
+    @assert 1 <= dims <= length(size(A))
+    if (length(size(A)) >= 2)
```
willow-ahrens, line 61: `length(size(A))` is more simply stated as `ndims(A)`

src/interface/compare.jl
```diff
@@ -55,3 +55,69 @@ end
...
+        return map(x -> x[2], reduce(minby, map(Pair, A, CartesianIndices(size(A))), dims=dims, init=Inf=>CartesianIndex(fill(0, length(size(A)))...)))
```
willow-ahrens, line 62: When we define these high-level functions, the pattern is to first define the function for lazy arguments in lazy.jl: https://github.com/finch-tensor/Finch.jl/blob/05f4a3d66cf51af30569715025daeb5d95f4efa5/src/interface/lazy.jl#L526

then define the eager version in terms of the lazy one in eager.jl: https://github.com/finch-tensor/Finch.jl/blob/05f4a3d66cf51af30569715025daeb5d95f4efa5/src/interface/eager.jl#L170-L172

src/interface/compare.jl
```diff
@@ -55,3 +55,69 @@ end
...
+    else
+        return map(x -> x[2], reduce(minby, map(Pair, A, 1:size(A)[1]), dims=dims, init=Inf=>0))
+    end
+end
+
+function helper_argmin(A)
```
willow-ahrens, line 68: could this be defined in terms of the previous function? i.e. `helper_argmin(A) = helper_argmin(A, (1:ndims(A)...,))`?

willow-ahrens (changes requested):
This looks great to me! Please also add:
1. tests. You can simply test
```
A = rand(4)
@test argmin(A) == argmin(Tensor(A))
...
```

Please also add a function `argmin_python` which follows the similar, but different, spec outlined in https://data-apis.org/array-api/latest/API_specification/generated/array_api.argmin.html#argmin

willow-ahrens:
You will need the shape of the lazy array for the strategy with eager.jl and lazy.jl to work. I'll add shapes for lazy arrays today, please in the meantime start by implementing tests and the `argmin_python` function.

01110011011101010110010001101111:
Where do you recommend putting the tests, @willow-ahrens?

willow-ahrens:
let's put them in interface_issues for now

src/interface/compare.jl
```diff
@@ -55,3 +55,78 @@ end
...
+function helper_argmin(A, dims)
+    # @assert 1 <= max(dims) <= ndims(A) && 1 <= min(dims) <= ndims(A)
+    if (ndims(A) >= 2)
+        return map(x -> x[2], reduce(minby, map(Pair, A, CartesianIndices(size(A))), dims=dims, init=Inf=>CartesianIndex(fill(0, length(size(A)))...)))
+    else
+        return map(x -> x[2], reduce(minby, map(Pair, A, 1:size(A)[1]), dims=dims, init=Inf=>0))
+    end
+end
+
+function argmin_python(A; dims, keepdims)
```
01110011011101010110010001101111, line 68: TODO: 
- [ ] Flattened
- [ ] Single Index always
- [ ] Look closer at the things returned in the reference

src/interface/lazy.jl
```diff
@@ -709,10 +709,33 @@ function compute_parse(ctx, args::Tuple)
...
+                map(
+                    Pair, 
+                    A, 
+                    CartesianIndices(size(A))), 
+                    dims=dims, 
+                    init=-Inf=>CartesianIndex(fill(0, length(size(A)))...)
+            )
+        ))
+        # NOTE: CURRENTLY DOES NOT WORK
+        B = expanddims(A1, dims)
```
01110011011101010110010001101111, line 733: There's a really odd bug where this does not work in the Lazy function. (It's called and updates the matrix but then it reverts when returned for some reason?) It does work when I pull this into the non-lazy version. Not entirely sure what's up here.

src/interface/compare.jl
01110011011101010110010001101111: https://data-apis.org/array-api/latest/API_specification/data_types.html#default-data-types

01110011011101010110010001101111: Tuple should work with multi-dimensional indexing (is the default return type): https://data-apis.org/array-api/latest/API_specification/indexing.html#multi-axis-indexing

willow-ahrens:
@kylebd99 I'd appreciate your help with the galley failures, I think it's just due to unrecognized ops. I'll investigate further but do let me know if there's something obvious

src/Galley/TensorStats/cost-estimates.jl
```diff
@@ -15,22 +15,24 @@ const SparseAllocateCost = 60
...
         elseif length(rel_conjuncts) == 0
-            new_def = merge_tensor_def(+, [get_def(stat) for stat in rel_disjuncts]...)
-            merge_tensor_stats_union(+, new_def, rel_disjuncts...)
+            new_def = merge_tensor_def(|, [get_def(stat) for stat in rel_disjuncts]...)
+            merge_tensor_stats_union(|, new_def, rel_disjuncts...)
         else
-            disjunct_def = merge_tensor_def(+, [get_def(stat) for stat in rel_disjuncts]...)
-            disjunct_stats = merge_tensor_stats_union(+, disjunct_def, rel_disjuncts...)
+            disjunct_def = merge_tensor_def(|, [get_def(stat) for stat in rel_disjuncts]...)
+            disjunct_stats = merge_tensor_stats_union(|, disjunct_def, rel_disjuncts...)
```
willow-ahrens, line 31: @kylebd99 No worries, I think I found the issue. When we pun on the tensor stats interface to compute lookup costs with +, we need to change the fill value in the stats objects to work with +. It worked so far because things are usually numeric.

willow-ahrens (approved):
Hi Shreya, great work on this! I just finished up by fixing bugs and adding a few tweaks to get the right semantics for the python versions.

01110011011101010110010001101111:
Cool, was reviewing the changes and noticed that a temp file I used was committed -- just removed that!

src/Galley/PhysicalOptimizer/protocol-selector.jl
```diff
@@ -93,25 +93,25 @@ function modify_protocols!(expr)
...
                 if length(indices_before_var) > 0
                     size_before_var = estimate_nnz(
                         reduce_tensor_stats(
-                            min,
-                            typemax(get_default_value(input)),
+                            |,
+                            false,
                             setdiff(get_index_set(input), indices_before_var),
-                            input,
+                            set_fill_value!(input, false),
```
willow-ahrens, line 105: @kylebd99 Here too, it seems the issue was not arbitrary ops, but arbitrary fill values that didn't get changed

PR #709, opened 2025-03-10
Change type signature for `Base.:+` square args

mtsokol:
Hi @willow-ahrens,

I noticed that for:

```julia
using Finch
using LinearAlgebra

A_ref = [1 2 0 4 0; 0 -2 1 0 1]
A = Tensor(Dense(SparseList(Element(0))), A_ref)
norm(A, 1) == norm(A_ref, 1)
norm(A, 2) == norm(A_ref, 2)  # fails
norm(A, 3) == norm(A_ref, 3)
```

the square option fails with an error:
```julia
ERROR: MethodError: no method matching +(::Finch.Square{Int64, Float64}, ::Finch.Square{Float64, Float64})

Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:587
  +(::T, ::T) where T<:Finch.Square
   @ Finch ~/Finch.jl/src/interface/lazy.jl:424
```
This PR fixes it.

### DIFF
--- a/src/interface/lazy.jl
+++ b/src/interface/lazy.jl
@@ -421,7 +421,7 @@ function Base.convert(T::Type{<:Number}, x::Square)
     return convert(T, root(x))
 end
 
-@inline function Base.:+(x::T, y::T) where {T<:Square}
+@inline function Base.:+(x::Square, y::Square)
     if x.scale < y.scale
         (x, y) = (y, x)
     end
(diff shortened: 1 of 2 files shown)


mtsokol:
Ok, looking at the CI I don't think my fix is the right one. But the issue described here is still relevant for computing vector norms with `p=2`.

PR #714, opened 2025-03-21
Release 1.2.2

mtsokol:
@willow-ahrens Thanks for #713! Can we have a new release?


willow-ahrens:
You should be all set! We're just waiting for the new version to work it's way through the Julia registry

PR #715, opened 2025-03-23
Implement interface for parallel scheduler

Paramuths:
Implement interface for parallel scheduler by 
- adding scheduler field in ParallelDimension struct
- implementing Scheduler enum

* Note there might be a potential problem with virtual representation of CPU: either literal(CPU) or VirtualCPU


willow-ahrens:
I've made some updates to the way we handle dimensions, so this PR should be easier to move forward with now. Notice that there is now a distinction between `ParallelDimension` (runtime struct), and `VirtualParallelDimension` (compile-time representation of the runtime struct). We will need to add a concrete and virtual version of each scheduler, rather than using an Enum, because at compile time we can't tell the difference between each scheduler type. I'll make a commit to do this so you can see how it works.

willow-ahrens:
Actually, if you have time, could you make a commit to move from an enum to separate structs that get virtualized (just like the paralleldimension)? I'm a little inundated with revisions to the Finch paper

PR #721, opened 2025-04-02
Implement interface for parallel scheduler

### LINKED ISSUE #607: Extensible Load Balancing
It would be great if we could support more configurable load balancing in Finch. Currently, we only statically partition.
1. Add an interface to CPUDevice that selects between :dynamic and :static strategies to @threads. https://github.com/willow-ahrens/Finch.jl/blob/a7f5b899d49051167aebd7568a65c2840ea6db69/src/architecture.jl#L35 https://github.com/willow-ahrens/Finch.jl/blob/a7f5b899d49051167aebd7568a65c2840ea6db69/src/lower.jl#L317
2. Generalize the parallelism model to configure chunk size and overpartition loop iterations. https://github.com/willow-ahrens/Finch.jl/blob/a7f5b899d49051167aebd7568a65c2840ea6db69/src/lower.jl#L292

willow-ahrens:
fixes #607


src/lower.jl
```diff
@@ -404,3 +411,98 @@ function lower_parallel_loop(ctx, root, ext::VirtualParallelDimension, device::V
...
+                            distribute(ctx_5, resolve(ctx_5, tns), subtask, diff, style),
+                        )
+                    end
+                    body = redistribute(ctx_5, body, diff)
+                    i = index(freshen(ctx, :i))
+                    root_2 = loop(i, VirtualExtent(tid, tid),
+                        loop(root.idx, ext.ext,
+                            sieve(
+                                access(
+                                    VirtualChunkMask(getstop(ext.ext), schedule.chk),
```
willow-ahrens, line 494: There's a big problem with using `VirtualChunkMask` like this: We don't know how many chunks there will be, and that `@threads for i = 1:nthreads` loop won't iterate over all the chunks (just the first `nthreads` chunks). 

In Julia, when we request `@threads :dynamic for i=1:n`, internally the scheduler picks a grain size and run the grains on e.g. `nthreads() * 4` workers. This is a challenge because Finch requires separately allocated memory spaces corresponding to each thread ID. For example, assume `nthreads() = 4` and we have allocated 4 scratch spaces, one for each thread. Then we call `@threads :dynamic for i = 1:100`, and we get 16 simultaneous tasks. How can we assign scratchspaces to tasks without conflict? If we wish to use a grain size of 4, we can't efficiently allocate 4/n separate scratchspaces.

Using Julia's `:dynamic` scheduler is also a problem because we may wish to parallelize along a sparse dimension. Imagine that the default scheduler wishes to process the range `i = 20:40`, which contains only one nonzero. The scheduler has no way to request to run a range of indices, so it would instead process each iteration independently.

Here's my recommendation:

1. Switch from VirtualChunkMask to a simple VirtualBandMask that iterates over a single chunk, and calculate the loop bounds yourself. Make this work for the static scheduler.
2. First try this approach: When it comes time to implement the dynamic scheduler, either use an atomic counter or a [`Channel`](https://docs.julialang.org/en/v1/manual/asynchronous-programming/#Communicating-with-Channels) to divvy up the chunks among workers. Then, the loop would look like:
```
nchunks = 0
@threads :static for tid = 1:cpu.n
    mychunk = (@atomic nchunks += 1)
    @finch for i = mychunk*chunksize : (mychunk + 1) * chunksize
        ...do work...
    end
end
```
3. Next, try the following, different approach: Use a `dynamic` schedule to divide the iterations, and use a `Channel` to [allocate different `tid` to different chunks](https://juliafolds2.github.io/OhMyThreads.jl/stable/literate/tls/tls/#The-safe-way:-Channel). The loop would look like:
```
tids = Channel{Int}(cpu.n)
for tid = 1:cpu.n
    put!(tids, tid)
end
@threads :greedy for mychunk = 1:nchunks
    tid = take!(tids)
    @finch for i = mychunk*chunksize : (mychunk + 1) * chunksize
        ...do work...
    end
    put!(tids, tid)
end
```

Benchmark both, I think they both may have uses. I think that the latter approach may lend itself better to nested parallelism.

willow-ahrens (commented):
There's some big changes we'll need to make to how we think about parallelism. I've sketched out in my comment some possible approaches we can take to dynamic scheduling.

PR #728, opened 2025-04-10
fixes #726

### LINKED ISSUE #726: Expandddims broken when called on the lazy result of an reduce
line https://github.com/finch-tensor/Finch.jl/blob/216ec6994620cb2ccd8509c536bd9a27fe0541c0/src/scheduler/optimize.jl#L168 strikes again

willow-ahrens:
fixes #726 by refusing to move extruding reorders into aggregates.

A more complete fix might also include a pass to calculate which dimensions are extruded. This PR includes such a pass, but does not use it or test it. We do leave a note about the semantics of mapjoin, which is that dimensions added through reorders are known as "extruded" dimensions, and these dimensions have length 1, but may be mapjoined with regular dimensions through broadcasting.


willow-ahrens:
@kylebd99 I thought you might find this of note.

PR #731, opened 2025-04-15
Parallel scheduler

Paramuths:
Implement parallel schedulers

fixes https://github.com/finch-tensor/Finch.jl/issues/607

formerly https://github.com/finch-tensor/Finch.jl/pull/721


src/dimensions.jl
```diff
@@ -208,45 +208,192 @@ function checklim(ctx::AbstractCompiler, a::FinchNode, b::FinchNode)
     end
 end
 
-@kwdef struct ParallelDimension{Ext,Device} <: AbstractExtent
+abstract type AbstractSchedule end
+
+abstract type AbstractVirtualSchedule end
+
+struct ZeroSchedule <: AbstractSchedule end
```
willow-ahrens, line 215: We need to rename the schedules:

ZeroSchedule -> FinchStaticSchedule, static_schedule()
OneSchedule -> is broken plz remove
TwoSchedule -> FinchGreedySchedule, greedy_schedule()
ThreeSchedule -> FinchJuliaSchedule, julia_schedule()

src/dimensions.jl
```diff
@@ -208,45 +208,192 @@ function checklim(ctx::AbstractCompiler, a::FinchNode, b::FinchNode)
     end
 end
 
-@kwdef struct ParallelDimension{Ext,Device} <: AbstractExtent
+abstract type AbstractSchedule end
```
willow-ahrens, line 211: I think we should move the schedule definititions to architecture.

willow-ahrens (changes requested):
Looks great! Just a few aesthetic changes and we can merge

src/dimensions.jl
```diff
@@ -208,45 +208,192 @@ function checklim(ctx::AbstractCompiler, a::FinchNode, b::FinchNode)
...
+    push_preamble!(
+        ctx,
+        quote
+            $chk_2 = $(ctx(chk))
+        end,
+    )
+    VirtualOneSchedule(value(chk_2, Int))
+end
+
+struct TwoSchedule <: AbstractSchedule
```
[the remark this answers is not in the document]
willow-ahrens: when you lower it, you probably need to use a QuoteNode.

for example

```
virtualize(ex, ::Type{TwoSchedule{schedule}}) where {schedule} =
   VirtualTwoSchedule(schedule, value(:($ex.chk), Int))

lower(ctx, node::VirtualTwoSchedule) = :(TwoSchedule{$(QuoteNode(node.schedule))}($node.chk))

function virtual_parallel_region(...)
    ...
    quote
        Thread.@threads $(QuoteNode(node.schedule)) for i ...
     

```

Paramuths:
> Looks great! Just a few aesthetic changes and we can merge

Fixed

src/architecture.jl
```diff
@@ -484,20 +497,250 @@ for T in [
...
+        subtask = VirtualCPUThread(value(tid, Int), device, ctx_2.code.task)
+        contain(ctx_2; task=subtask) do ctx_3
+            f(ctx_3, i_lo, i_hi)
+        end
+    end
+
+    return quote
+        Threads.@threads $(QuoteNode(schedule.schedule)) for $tid in 1:($(ctx(device.n)))
+            Finch.@barrier begin
+                @inbounds @fastmath begin
```
willow-ahrens, line 684: How can we call distribute_device here, instead of inside the while loop?

willow-ahrens (changes requested):
A single concern: We need to call `distribute_device` outside of the while loop to realize the benefits of this scheduler.

willow-ahrens:
Very clean fix! Next order of business is passing CI

willow-ahrens (approved):

willow-ahrens:
To get CI to pass, we may need to change the schedule argument in julia 1.10.7. Perhaps you could check which julia version we're on and only use :greedy in 1.11?

PR #734, opened 2025-04-16
Wma/keepdims

### LINKED ISSUE #727: We need to call expanddims after every reduce call to match standard julia behavior.
By default, Base julia reductions keep dims, unless you pass `:` in which case they drop all the dims. In Python, there's a keepdims argument which can be set, but in julia we don't have this. Instead, you could fuse a dropdims into the reduction.

willow-ahrens:
fixes #727

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Finch"
 uuid = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
 authors = ["Willow Ahrens"]
-version = "1.2.3"
+version = "1.2.4"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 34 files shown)


willow-ahrens:
@kylebd99 there appears to be some non-determinism and/or julia-version-specific behavior in galley, uncovered by this PR, could you take a look?

kylebd99:
Things that I tried to stabilize Galley's windows performance:
1. Change all Dicts to OrderedDicts
2. Change StaticBitset to Bitset.jl
3. Change UInt128 to Float64 in DCStat Struct

So far, no dice :/

willow-ahrens:
I'll take a look, and if I don't see anything I'll remove it from the test suite.

willow-ahrens:
Thanks for your help!

kylebd99:
Of course, happy to help whenever! FWIW, I'm also comforted by the fact that the non-deterministic behavior is only occurring on the 1x1 tensor example.

willow-ahrens:
@kylebd99 after changing Set to OrderedSet, it looks like the two systems are in agreement.

willow-ahrens:
a tricky bug I found: ordered sets seem to have different hashes when the orders are different. I'll file an issue in the datastructures library.

PR #735, opened 2025-04-18
Replace OrderedSet with SortedSet to avoid hash bug

kylebd99:
I'm just a bit paranoid about the existence of a hash bug in OrderedSet, so for the sake of prudence I suggest we use SortedSet. My guess is that the performance impact will be fairly negligible.


willow-ahrens:
does this work? I tried this but realized we needed to hash the keys, which is a little fraught. We could use `StableHash`, but it keeps references to everything so the whole code would need to be modified to accept a limited-lifetime hash function object.

willow-ahrens:
You're right that this is a problem though, because galley tests now fail on the finch-tensor-python. A fix is proposed in https://github.com/JuliaCollections/OrderedCollections.jl/pull/149, which we could pull in for now?

kylebd99:
That works for me. Let's just pull in the fix.

willow-ahrens:
I've got an alternate approach to just use Set when it's the key of a dictionary: https://github.com/finch-tensor/Finch.jl/pull/736

willow-ahrens:
I think the orderset fix is actaully wrong. objects in julia are hashed based on isequal, not equal. I'll work on a different fix for this.

PR #736, opened 2025-04-18
Wma/quickfix 1.2.5

willow-ahrens:


kylebd99:
I'd prefer a uniform choice of Sets, but I get that having a dependency on your fork of OrderedCollections is also bad. So, happy with either approach.

PR #737, opened 2025-04-18
Avoid OrderedSet hash bug

willow-ahrens:


kylebd99:
I think the issue here is due to the hashing of the Set type. The standard set only hashes the elements:

```
julia> a = Set{Any}([1,2])
Set{Any} with 2 elements:
  2
  1
julia> b = Set{Int}([1,2])
Set{Int64} with 2 elements:
  2
  1
julia> hash(a)
0x78b09051102fa69e
julia> hash(b)
0x78b09051102fa69e
```

kylebd99:
I'll go ahead and make the change real quick.

willow-ahrens:
Ah, I see why that would be an issue! If this passes tests, let's merge it. I have one I forgot to push that passes tests in https://github.com/finch-tensor/Finch.jl/pull/738 by just constructing a Set and hashing that which we can use as a backup.

willow-ahrens:
Actually, looking at the code it seems that I did indeed push that fix. I had put Stable set in util/convenience.jl (maybe a weird place for it), and you had put it in stable_set.jl, which I don't think is getting included. Either way, I tried the python tests locally and they seemed okay, so I'm not sure what's going on.

willow-ahrens:
I'll let you take over on this one, I feel that I'm out of my depth as to why the python tests are failing on CI. They are failing because they return an incorrect dimension size for the tensor now.

kylebd99:
Sounds good, I'll see if I can sort this out today. I think the convenience_utils version is just missing the fixed hash.

PR #744, opened 2025-04-22
Make qos_fill and qos_stop local for each thread in sparsebytemap

Paramuths:


Project.toml
```diff
@@ -63,4 +63,4 @@ julia = "1.10"
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 NPZ = "15e1cf62-19b3-5cfa-8e77-841668bca605"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-TensorMarket = "8b7d4fe7-0b45-4d0d-9dd8-5cc9b23b4b77"
\ No newline at end of file
+TensorMarket = "8b7d4fe7-0b45-4d0d-9dd8-5cc9b23b4b77"
```
willow-ahrens, line 66: Do we need this dep? Let's remove it if we can

src/tensors/levels/sparse_bytemap_levels.jl
```diff
@@ -254,8 +254,8 @@ function distribute_level(
         distribute_buffer(ctx, lvl.tbl, arch, style),
         distribute_buffer(ctx, lvl.srt, arch, style),
         lvl.shape,
-        lvl.qos_fill,
-        lvl.qos_stop,
+        distribute_buffer(ctx, lvl.qos_fill, arch, style),
```
willow-ahrens, line 257: I didn't realize this was such a simple fix. Thank you so much!

willow-ahrens (changes requested):
Remove that dep from project.toml and this is good to go!

willow-ahrens (approved):
nevermind about the tensormarket dep, I read the diff wrong

PR #748, opened 2025-05-07
MAINT: Tweak `push_fields` pass

mtsokol:
Hi @willow-ahrens,

I might be wrong here but I think that `reorder/aggregate` rule in `push_fields` pass is never used as `¬issubset(A ∩ B, B)` is always false regardless of inputs.

### DIFF
--- a/src/FinchLogic/nodes.jl
+++ b/src/FinchLogic/nodes.jl
@@ -59,6 +59,8 @@ table
 
 Logical AST expression for mapping the function `op` across `args...`.
 The order of fields in the mapjoin is `unique(vcat(map(getfields, args)...))`
+Dimensions of fields from different arguments must match, and fields which are
+missing from an argument are broadcasted.
 """
 mapjoin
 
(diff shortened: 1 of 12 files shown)


willow-ahrens:
notice that we use "issubsequence", not "issubset". I think the intention is to push a reorder through an aggregate when the fields of the reorder are in a different permutation than the fields of the argument, ignoring dropped or expanded dimensions. I'm not sure if we really should care to do this though, because in general we don't usually push reorders through aggregates. Let's wait for https://github.com/finch-tensor/Finch.jl/pull/750 to merge, and benchmark this change. I think the intention is to try to avoid permuting the output of an aggregate, if possible.

mtsokol:
Thanks for an explanation - I'm still a bit confused by the `if` statement here. One thing would help me: Can you provide some exemplary values here for `arg` and `idxs_2` os that the statement below evaluates to true?
```julia
!issubsequence(intersect(getfields(arg), idxs_2), idxs_2)
```

willow-ahrens (approved):
I'm happy to remove this rule if it passes tests and causes no regressions.

mtsokol:
> I'm happy to remove this rule if it passes tests and causes no regressions.

Let's first release what we have currently in the `main` branch (we'll be sure we have latest release without regression).
This weak can wait - I will test it separately, it isn't a priority.

PR #750, opened 2025-05-13
fix push fields

willow-ahrens:
fixes https://github.com/finch-tensor/Finch.jl/issues/749


willow-ahrens:
Hi @mtsokol, sorry it took a while (I was chasing down bugs). This PR fixes the autoscheduler to maintain consistent FinchLogic semantics throughout the optimization process.

PR #753, opened 2025-05-16
first pass at a fix tor triangle counting performance

willow-ahrens:
Here's my fix for triangle counting. I think it should resolve performance issues.


willow-ahrens:
I removed the offending test `reduce(A, dims=(1, 2))`. I know this is sort of bad, but this isn't our biggest problem right now. If someone has time they can sort it out later. I left a TODO in the code, but I need to move on to other problems.

willow-ahrens:
Feel free to merge this when you're ready @mtsokol

mtsokol:
Hi @willow-ahrens, I reran benchmarks with this patch and the regression is fixed now, thank you! Let's also make a new release with that patch, I'll ship it to `finch-tensor-python` and `pydata/sparse` and this can be the version for PLDI presentation.

PR #756, opened 2025-05-18
Remove OpenSSL compat dep

mtsokol:
Looks like https://github.com/JuliaPy/CondaPkg.jl/issues/181 was fixed, so I think we can drop version pin (especially that macos and windows give unsatisfiable requirements error now).


mtsokol:
It's going to be another release today but I think it's needed for https://github.com/finch-tensor/finch-tensor-python/actions/runs/15095457137/job/42429827352?pr=120.

PR #760, opened 2025-06-05
Update README.md

kylebd99:
With Galley being in this repo, seems like it should be in the learn more section. Also, I'm linking to this repo for the Galley code, so I don't want to confuse folks.


willow-ahrens (commented):
Should we use the official DOI?

PR #770, opened 2025-10-14
Use `arg` in `set_loop_order` instead of `A`

mtsokol:


willow-ahrens (approved):
I'll merge my 1.12 fixes into this, so we don't need to constrain julia version

PR #785, opened 2025-11-19
CPU Tagging, SPList, SPDict Bugfixes

AdrianGushin:
This pull request implements CPU tagging for Finch's main branch. It also fixes some bugs that occur in the sparse list and dict levels when interacting with parallelism.


src/architecture.jl
```diff
@@ -96,15 +96,15 @@ A device that represents a serial CPU execution.
 """
 struct Serial <: AbstractTask end
 serial() = Serial()
-get_device(::Serial) = CPU(1)
+get_device(::Serial, id=1) = CPU(id, 1)
```
willow-ahrens: can we default the id to a symbol like `:default` to avoid using integers, which have confusion with the number of cores (which we probably don't want here)?

PR #799, opened 2026-03-18
Support Matrix Market symmetry and binsparse fixtures

BenBrock:
## Summary

This PR makes a few updates and modifications to Finch.jl to support the Binsparse files and Matrix Market files used by the (WIP) [Binsparse testing framework](https://github.com/BenBrock/binsparse-tests).

This consists primarily of 1) routing `.mtx` files to `MatrixMarket.jl` instead of TensorMarket, since TensorMarket fails on Matrix Market files; 2) Adding support for `iso[...]` scalar values; and 3) extending Finch's Binsparse reader to properly handled structured matrices (symmetric, etc.).

**The rest of this PR summary is generated by Codex.  Most of the code changes in this PR are AI-generated and need review (thus it's a draft PR).**

The main changes are:

- add native .mtx support through MatrixMarket.jl instead of routing .mtx files through the TensorMarket path
- add a MatrixMarketExt extension that reads Matrix Market files into Finch tensors and writes Finch tensors back to Matrix Market
- extend binsparse reading to support iso[...] value encodings
- extend binsparse reading to honor matrix structure metadata for symmetric, skew_symmetric, and hermitian matrices
- add a COO-specific binsparse matrix read path for 2D matrix fixtures
- fix COO index ordering in binsparse read/write so generated files line up with the current fixture layout

## Motivation

The current Finch fileio path had a few gaps that showed up immediately when running the new binsparse parser test framework:

- symmetric Matrix Market files were not handled correctly through the normal .mtx path
- binsparse fixtures using iso[...] value metadata could not be read
- binsparse fixtures with structure metadata such as symmetric_lower could not be reconstructed correctly

This branch closes those gaps so Finch can participate in the same roundtrip fixture tests as the other parser implementations.

## Tests

Added coverage for:

- reading binsparse HDF5 fixtures with iso[bint8] values and symmetric_lower structure
- reading and writing symmetric Matrix Market files

Validated with:

- Finch fileio test suite: 109 passed
- binsparse interoperability harness against the local Finch adapter: 4 passed

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Finch"
 uuid = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
-authors = ["Willow Ahrens"]
 version = "1.2.12"
+authors = ["Willow Ahrens"]
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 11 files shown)


BenBrock:
@willow-ahrens Can you comment in the MatrixMarket.jl dependency here?  Looks like TensorMarket fails to read certain Matrix Market files directly.

The rest of the changes are to reading Binsparse files, which I think need to be made.

PR #805, opened 2026-07-07
buffer-first datastructures

willow-ahrens:


src/tensors/levels/coalesce_levels.jl
```diff
@@ -34,13 +34,15 @@ function CoalesceLevel(device::Device, lvl::Lvl) where {Device,Lvl}
     while typeof(coal_lvl) <: CoalesceLevel
         coal_lvl = coal_lvl.lvl
     end
+    P = get_num_tasks(device)
     coalescent = similar_level(
         coal_lvl, level_fill_value(Lvl), level_eltype(Lvl), level_size(lvl)...
     )
+    coalescent = coalesce_similar_level(coalescent)
```
AdrianGushin: bit confused what this line is doing, can't seem to find the function definition. i thought the coalescent was already defined on line 38

src/tensors/levels/sparse_bytemap_levels.jl
```diff
@@ -626,128 +663,155 @@ end
...
-            if nz_id >= length(srt[proc_id]) && proc_id < P
-                proc_id += 1
+    seen = sum(length, chunk_srt)
+    resize!(lvl_srt, seen)
+    global_fbr_map = Vector{Int}(undef, nnz)
+    local_fbr_map = Vector{Int}(undef, nnz)
+    task_map = Vector{Int}(undef, nnz)
+    q_offset = 1
+    map_offset = 1
+    @inbounds for part in 1:P
```
AdrianGushin: could these copy-to's be slow? not sure if map_count could be large but it seems like it could be. maybe this needs to be done in parallel?

AdrianGushin (commented):
i left two questions. i like the new hash level.

src/tensors/levels/sparse_bytemap_levels.jl
willow-ahrens: I didn't focus on optimizing this as much. I was thinking I'd wait for desc interface to do performance optimization. I'm going to tag a "wingspan" release so that these changes don't slow down your code.

willow-ahrens: https://github.com/finch-tensor/Finch.jl/releases/tag/wingspan-arxiv-v0

Issue #708, opened 2025-03-10
Error when computing negative and finite vector norm


mtsokol:
Hi @willow-ahrens,

I started porting vector norm and I get a `RuleRewriteError` for all negative values of `p`:

```julia
using Finch
using LinearAlgebra

A_ref = [1 -2 -3; 2 3 -1]
A = Tensor(Dense(SparseList(Element(0.0))), A_ref)

norm(A_ref, 1)
norm(A, 1)
norm(A_ref, Inf)
norm(A, Inf)
norm(A_ref, -Inf)
norm(A, -Inf)
norm(A_ref, -1)
norm(A, -1)  # The last line fails
```

Here's error log:

<details>

```julia
julia> norm(A, -1)
ERROR: RewriteTools.RuleRewriteError(call(~f, ~a..., ~b, ~c, ~d...) => if isidentity(alg, f, b)
        #= /home/mtsokol/Finch.jl/src/symbolic/simplify.jl:59 =#
        call(f, a..., c, d...)
    end, call(literal(*), literal(Finch.Power{Float64, Float64, Int64}(Inf, 0.0, -1)), cached(call(literal(+), call(literal(-), value(i0_start_2, Any)), value(phase_stop_2, Any)), literal(call(literal(+), literal(1), call(literal(-), value(i0_start_2, Any)), call(literal(-), cached(value(A0_lvl_2_i, Any), literal(call(literal(max), value(A0_lvl_2_i, Any), call(literal(+), value(i0_start_2, Any), literal(1))))), literal(1)))))))
Stacktrace:
   [1] (::RewriteTools.Rule{…})(term::Finch.FinchNotation.FinchNode)
     @ RewriteTools ~/.conda/envs/finch-jl-dev/share/julia/packages/RewriteTools/CB1M6/src/rule.jl:151
   [2] (::RewriteTools.Rewriters.Chain{Vector{RewriteTools.Rule{RewriteTools.Term}}})(x::Finch.FinchNotation.FinchNode)
     @ RewriteTools.Rewriters ~/.conda/envs/finch-jl-dev/share/julia/packages/RewriteTools/CB1M6/src/rewriters.jl:164
   [3] (::RewriteTools.Rewriters.Fixpoint{RewriteTools.Rewriters.Chain{Vector{RewriteTools.Rule{…}}}})(x::Finch.FinchNotation.FinchNode)
[...]

Issue #686, opened 2025-01-03
- densifies zeros


willow-ahrens:
A - A results in all the zero entries getting converted to -0.0 and being stored explicitly 😬 

```
julia> A = fsprand(5, 5, 0.5)
5×5 Tensor{SparseCOOLevel{2, Tuple{Int64, Int64}, Vector{Int64}, Tuple{Vector{Int64}, Vector{Int64}}, ElementLevel{0.0, Float64, Int64, Vector{Float64}}}}:
 0.0       0.836373  0.73536  0.0         0.83067
 0.0       0.0       0.0      0.0         0.390398
 0.677004  0.0       0.0      0.0         0.605428
 0.0       0.0       0.0      0.00907705  0.728524
 0.0       0.0       0.0      0.878594    0.0

julia> A - A
5×5 Tensor{SparseDictLevel{Int64, Vector{Int64}, Vector{Int64}, Vector{Int64}, Dict{Tuple{Int64, Int64}, Int64}, Vector{Int64}, SparseDictLevel{Int64, Vector{Int64}, Vector{Int64}, Vector{Int64}, Dict{Tuple{Int64, Int64}, Int64}, Vector{Int64}, ElementLevel{0.0, Float64, Int64, Vector{Float64}}}}}:
 -0.0   0.0   0.0  -0.0   0.0
 -0.0  -0.0  -0.0  -0.0   0.0
  0.0  -0.0  -0.0  -0.0   0.0
 -0.0  -0.0  -0.0   0.0   0.0
 -0.0  -0.0  -0.0   0.0  -0.0
 
 julia> countstored(A - A)
25
```

willow-ahrens:
After #383, the handling of `-0.0` has proven to be a continuing issue. The problem is that there's two ways to evaluate` 0.0 + (-0.0)` in Finch. We could follow floating point semantics, which gives `0.0`. We could also use the fact that `0.0` is an annihilator, which gives `-0.0`. I'm beginning to think the right choice may be to defer to actual evaluation semantics whenever possible.

willow-ahrens:
After some thought, we can summarize this issue as:
- `-0.0 + 0.0 = 0.0` (from floating point semantics)
- `-0.0 + 0.0 = -0.0` (from `x + 0 = x` rule)
We need one of these to win out over the other when we do constant propagation in Finch.
[...]

PR #810, opened 2026-08-20
fix: `norm(A, 2)` returned a complex number for complex tensors (#809)

### LINKED ISSUE #809: [BUG] norm(A, 2) returns a complex number for complex tensors
`norm(A, 2)` on a complex tensor returns a **complex** value. A norm is real by definition.

```julia
julia> using Finch, LinearAlgebra

julia> A = Tensor(ComplexF64[1+2im, 3-1im]);

julia> norm(A)
2.27872385417085 - 0.43884211690225483im     # should be 3.872983346207417

julia> norm(ComplexF64[1+2im, 3-1im])        # LinearAlgebra, for comparison
3.872983346207417
```

## Cause

`square` in `src/interface/lazy.jl`:

```julia
@inline square(x) = Square(sign(x)^2 / one(x), norm(x))
```

For a complex `x`, `sign(x)` is the unit-modulus phase `x / |x|`, so `sign(x)^2` is a *phase*
rather than `1`. `Square(arg, scale)` represents `arg · scale²`, so the pair becomes
`(x/|x|)² · |x|² = x²`, and the reduction computes `sqrt(sum(x^2))` instead of
`sqrt(sum(abs2(x)))`.

That the output matches `sqrt(sum(x.^2))` to the last digit:

```julia
julia> sqrt(sum(ComplexF64[1+2im, 3-1im] .^ 2))
2.27872385417085 - 0.4388421169022545im
```

## Scope: `p == 2` only

Every other `p` is correct, because that branch pre-maps `norm` over the elements
(`map(power, map(norm, arr, p), p)`) so `power` only ever sees non-negative reals. The `p == 2`
branch applies `square` directly to the raw elements.

Measured on `ComplexF64[3+4im, 1-1im, 0]`:

| p | Finch | LinearAlgebra | |
|---|---|---|---|
| 2 | `2.836 + 3.879im` | `5.196152422706632` | ✗ |
| 1 | `6.414213562373095` | `6.414213562373095` | ✓ |
| Inf | `5.0` | `5.0` | ✓ |
| 0 | `2` | `2.0` | ✓ |
| -Inf | `0.0` | `0.0` | ✓ |
| 3 | `5.037431439849898` | `5.037431439849897` | ✓ |
| 0.5 | `11.732509459318084` | `11.732509459318083` | ✓ |

Real inputs are unaffected at every `p`.

## Suggested fix

`abs2(sign(x))` is `1` for every nonzero `x` — real or complex — and `0` at `x == 0`, which is what
the scale-factored sum needs, and it preserves the zero-handling that `sign(x)^2` provided:

```julia
@inline square(x) = Square(abs2(sign(x)) / one(real(x)), norm(x))
```

`power` has the same shape (`sign(x)^p`) and is correct today only because of the branch ordering
above; making it `abs(sign(x))^p` keeps it correct independently.

Environment: Finch v1.4.0, Julia 1.12.2. Happy to open a PR — I have the fix and a test that fails
without it.

sotashimozono:
Fixes #809.

```julia
julia> norm(Tensor(ComplexF64[1+2im, 3-1im]))
2.27872385417085 - 0.43884211690225483im     # should be 3.872983346207417
```

## Cause

`square` built its `Square` from `sign(x)^2`. For a complex `x`, `sign(x)` is the unit-modulus phase
`x / |x|`, so `sign(x)^2` is a *phase* rather than `1` — and since `Square(arg, scale)` represents
`arg · scale²`, the pair became `(x/|x|)² · |x|² == x²`. The reduction therefore computed
`sqrt(sum(x^2))` instead of `sqrt(sum(abs2(x)))`, which matches the observed output to the last
digit:

```julia
julia> sqrt(sum(ComplexF64[1+2im, 3-1im] .^ 2))
2.27872385417085 - 0.4388421169022545im
```

## Fix

```diff
-@inline square(x) = Square(sign(x)^2 / one(x), norm(x))
+@inline square(x) = Square(abs2(sign(x)) / one(real(x)), norm(x))
```

`abs2(sign(x))` is `1` for every nonzero `x` — real or complex — and `0` at `x == 0`, so it keeps
the zero-handling `sign(x)^2` provided while dropping the phase. `one(real(x))` preserves the float
promotion without dragging the divisor complex.

## Scope is `p == 2` only

The other branches pre-map `norm` over the elements (`map(power, map(norm, arr, p), p)`), so they
only ever see non-negative reals and were already correct. Only `p == 2` applies `square` to the raw
elements. Measured on `ComplexF64[3+4im, 1-1im, 0]`:

| p | before | LinearAlgebra | |
|---|---|---|---|
| 2 | `2.836 + 3.879im` | `5.196152422706632` | ✗ |
| 1 | `6.414213562373095` | `6.414213562373095` | ✓ |
| Inf | `5.0` | `5.0` | ✓ |
| 0 | `2` | `2.0` | ✓ |
| -Inf | `0.0` | `0.0` | ✓ |
| 3 | `5.037431439849898` | `5.037431439849897` | ✓ |
| 0.5 | `11.732509459318084` | `11.732509459318083` | ✓ |

Real inputs were unaffected at every `p`, and remain so.

`power` gets the same treatment (`abs(sign(x))^p`). It is correct today *purely* because of that
branch ordering — this makes the ordering not load-bearing. No behaviour change, since it only ever
receives non-negative reals.

## Test

Added to `interface_issues`: `ComplexF64` vector and matrix, `ComplexF32`, an all-zero complex
tensor, plus `Float64`/`Int` controls, at `p = -Inf, 0, 1, 2, 3, Inf`, asserting both the value and
`isreal`.

Checked it has teeth — reverting only the `src` change fails it on all three nonzero complex cases:

```
FAIL ComplexF64 p=2  got=2.27872385417085 - 0.43884211690225483im  want=3.872983346207417
FAIL ComplexF64 p=2  got=2.6073691484423227 + 4.218811903397548im  want=5.5677643628300215
FAIL ComplexF32 p=2  got=1.4142134f0 + 1.4142135f0im               want=2.0
```

## What I verified, and what I did not

- ✅ the new cases pass with the fix and fail without it
- ✅ formatting clean under your pinned JuliaFormatter 1.0.62 with `.JuliaFormatter.toml`'s settings
- ⚠️ **the full suite was still running when I opened this** — it is slow to compile locally and I
  did not want to hold the PR on it. CI here is the real check; happy to chase anything it turns up.

Found while evaluating Finch as a storage backend for ITensors (complex tensors throughout, so
`norm` is on the hot path for us).

Environment: Finch v1.4.0 @ 8bd2c92, Julia 1.12.2.

### DIFF
--- a/src/interface/lazy.jl
+++ b/src/interface/lazy.jl
@@ -571,7 +571,7 @@ end
 
 function LinearAlgebra.norm(arr::LazyTensor, p::Real=2)
     if p == 2
-        return map(root, sum(map(square, arr)))
+        return map(root, sum(map(square, map(norm, arr))))
     elseif p == 1
         return sum(map(abs, arr))
     elseif p == Inf
--- a/test/suites/interface_tests.jl
+++ b/test/suites/interface_tests.jl
@@ -955,6 +955,28 @@ end
                     end
                 end
 
+                # https://github.com/finch-tensor/Finch.jl/issues/809
+                let
+                    for A_ref in (
+                        ComplexF64[1 + 2im, 3 - 1im],
+                        ComplexF64[3+4im 1-1im; 0 -2im],
+                        ComplexF32[1 + 1im, -1 - 1im],
+                        ComplexF64[0, 0],
+                    )
+                        A = Tensor(A_ref)
+                        for p in [-Inf, 0, 1, 2, 3, Inf]
+                            @test norm(A, p) ≈ norm(A_ref, p)
+                            @test isreal(norm(A, p))
+                        end
+                    end
+                    for A_ref in (Float64[3, -4, 0], Int[1 2; 3 4])
+                        A = Tensor(A_ref)
+                        for p in [-Inf, 0, 1, 2, 3, Inf]
+                            @test norm(A, p) ≈ norm(A_ref, p)
+                        end
+                    end
+                end
+
                 # https://github.com/finch-tensor/Finch.jl/pull/709
                 let
                     A_ref = [1 2 0 4 0; 0 -2 1 0 1]


sotashimozono:
Force-pushed to strip explanatory comments out of `src` and the test — that reasoning belongs here
in the PR, not in your codebase, and I had let my own repo's (much more comment-heavy) house style
leak into someone else's.

`src/interface/lazy.jl` is now a two-line change:

```diff
-@inline square(x) = Square(sign(x)^2 / one(x), norm(x))
+@inline square(x) = Square(abs2(sign(x)) / one(real(x)), norm(x))

-@inline power(x, p) = Power(sign(x)^p, norm(x), p)
+@inline power(x, p) = Power(abs(sign(x))^p, norm(x), p)
```

and the test carries only the bare issue URL, matching the convention already used for #708 / #709 /
#686 in that file.

Re-verified after the strip: formatting clean under JuliaFormatter 1.0.62, and the new cases still
pass with the fix and fail without it.

